PR #463, opened 2021-12-08
feat(Button): add "kind" prop; deprecate "type" prop

### LINKED ISSUE #462: [Button] (breaking) Change `type` prop to `kind`
See: https://github.com/narmi/design_system/pull/458#pullrequestreview-826464850

`type` should be reserved for the native `type` attribute.
The `type` prop of `Button` should be renamed `kind`.

akdetrick:
fixes #462 

Button style variant controlled by `kind` prop instead of `type`. This is not a breaking change; `type` will be removed in a future major version ( issue #464 )

### DIFF
--- a/src/Button/index.js
+++ b/src/Button/index.js
@@ -14,7 +14,8 @@ import AsElement from "../util/AsElement";
  */
 const Button = ({
   disabled = false,
-  type = "primary",
+  type,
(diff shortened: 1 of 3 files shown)


akdetrick:
Doing this as an actual breaking change instead. Closing.

Issue #473, opened 2021-12-09
[Build,CI] pass artifact from install/build to other jobs if possible


akdetrick:
Each job has to do is own NPM install and full build currently. If there's a way to pass an artifact from an initial build job, things would be much cleaner and faster in github actions

akdetrick:
not worth the time right now

PR #478, opened 2021-12-09
ci: Use GitHub CodeQL Analysis

chris-griffin:
Closes https://github.com/narmi/design_system/issues/483

Enables the GitHub CodeQL Analysis bot. See https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql for more information

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,59 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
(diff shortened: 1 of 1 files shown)


akdetrick (approved):
Let's give this a shot - if PR checks take too long to run, we can always try speeding them up with #473 or just run this job on a schedule on `master`

akdetrick:
do we need to do anything with the results file? `/home/runner/work/design_system/results/javascript.sarif"`

Or can we let this live and die in the job that runs? I'm assuming if it finds anything, a useful error will be thrown and fail the job

PR #487, opened 2021-12-14
build(dependencies): downgrade react-test-renderer

akdetrick:
`react-test-renderer` at v17 requires a peer dependency on React 17. This project is at React 16, so we must downgrade  `react-test-renderer` to v16 to properly resolve the peer dependency.


akdetrick:
not sure if we actually need this package...

PR #497, opened 2021-12-16
build(dependencies): rm react-test-renderer; replace node-sass with sass

akdetrick:
- removes unused `react-test-renderer` dependency which declares a peer dep that messes up our dependency tree
- replaces `node-sass` with `sass`. `node-sass` uses `libsass` under the hood which does not yet have support for `arm64` architecture


graphicsforge (approved):
Yeah, that sounds reasonable!
Shouldn't include a testing dependency if we're not using it!

PR #500, opened 2021-12-20
fix(Button): remove width restriction from plain button

akdetrick:
Instead of setting a content max width on every button variant, we can set it only on the plain and secondary variants. Buttons of `plain` type should not have a restricted width.


xiexiewenting (approved):
tried it out on storybook, LGTM! thanks 🎉

PR #504, opened 2021-12-21
fix: sub-header style should display text color as --nds-black

agreatscott:
This style class has relatively little usage currently so the change should be pretty safe.


pr-narmi (approved):

xiexiewenting (approved):
LGTM

Issue #512, opened 2022-01-09
Skeleton state loading component


agreatscott:
### Short Description

A component that displays the loading state as a skeleton, to be used mainly while async operations are processing.

### Mockup

https://zeroheight.com/8ac87d4ba/p/03e21f-loading-states

### Responsibilities

**This component should be responsible for the following:**
- Provide a low fidelity representation of content before it appears on the page. 
- Be configurable as to the size of the content - display text, paragraph, single line, etc.

**This component should NOT be responsible for...**
- anything to do with the async operation or logic that's resulting in the need for the loading state

### Props Signature

```
<LoadingSkeleton
  isLoading={false}
  type="singleLine"
```

```
LoadingSkeleton.propTypes = {
  isLoading: PropTypes.bool,
  type: PropTypes.oneOf(["singleLine", "paragraph", "displayTextLarge" ...])
}
```

For the initial component we could perhaps just support paragraph or display text. For paragraph - it might also be good for the component to take a prop detailing number of lines?

PR #513, opened 2022-01-10
feat(LoadingSkeleton): add component

agreatscott:
close https://github.com/narmi/design_system/issues/512

Should fit the bill as per the design spec, but not sure I love shoving both the title mask and the paragraph mask into one component. My issue is that are prop combinations that have no effect: for example, `lines` and `title` only apply if `content` is set to `paragraph`, and `size` only applies if `content` is set to `displayText`, so maybe it makes sense to have two LoadingShadow components? 

Or I suppose, we could remove the `content` prop altogether, and allow the consumer to set the number of lines, the size of the lines, and whether there is a title row regardless of how certain combinations aren't strictly compatible with the Narmi style guide. 

Then again maybe I'm overthinking it, and this is fine, as long as the props are well documented.

### DIFF
--- a/src/LoadingSkeleton/index.js
+++ b/src/LoadingSkeleton/index.js
@@ -1,0 +1,40 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+/**
+ * Used to mask a limited block of loadable content. When `isLoading` is set to true,
+ * the content area will be be replaced by a shadowed, skeletal imitation of the
+ * pending content.
+ */
+
+ const ParagraphSkeleton = ({ lines=3, showTitle=true}) => (
+  <div className="nds-paragraph-skeleton">
+    {showTitle && <div className="nds-line-block header" />}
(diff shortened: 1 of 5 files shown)


src/LoadingSkeleton/index.js
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Only applies if `content` prop is set to `paragraph`.
+   * The number of lines to be shown by the skeleton.
+   */
+  lines: PropTypes.number,
+  /**
+   * Only applies if `content` prop is set to `paragraph`.
+   * When `true`, a skeletal title row will be shown above the first line.
+   */
+  title: PropTypes.bool,
```
akdetrick, line 52: I need to document some preferred naming conventions, but we should prefer an auxiliary verb for boolean props:

```diff
-title: PropTypes.bool,
+hasTitle: PropTypes.bool,
```

akdetrick (changes requested):
one minor naming thing

akdetrick (approved):

PR #517, opened 2022-01-11
style(reset): add reset and base CSS

### LINKED ISSUE #494: [Reset CSS,semantic-ui] - add base custom reset CSS
### Short Description

For components to be as portable as they need to be, NDS needs to be more opinionated about base element styles instead of relying on user agent defaults or base styles in the consumer.

Adding a custom reset on base elements also enables us to [override semantic-ui by changing CSS loading order].(https://github.com/narmi/banking/issues/13781):

The NDS base stylesheet must cover all of the element selectors from our builds of semantic-ui in addition to any sensible defaults we want to set.

#### ⚠️ proceed with caution and test often 🔬

Adding a global custom reset to NDS may have side effects in `banking`. Test for regressions using both load orders for semantic-ui and design_system stylesheets.

akdetrick:
fixes #494 

### What changed
- Adds `reset.css` with common styles to reset/normalize elements across browsers
- Adds `defaults.css` to set NDS default styles on some base elements
- Makes some minor style fixes to bugs exposed by adding base styling

### Why do we need this?
Setting foundational styles helps make NDS styling more portable and predictable across different browsers and consumers. The styles in `default.css` are also essential to overriding [underqualified selectors from semantic ui in banking](https://gist.github.com/akdetrick/72ee0385f10a5b4e14a1326e2a416acc)

---

Tested changes with `1.27.0-beta.0` in banking.

**This is not a high risk change**
`semantic-ui` still comes last in load order in `banking`, so it will override any selectors of the same specificity from NDS. The scarier, higher risk change is adjusting the load order in banking which will be handled by this ticket:
- https://github.com/narmi/banking/issues/13781

### DIFF
--- a/src/Button/index.scss
+++ b/src/Button/index.scss
@@ -1,6 +1,5 @@
 .nds-button,
 .nds-button.resetButton {
-  box-sizing: border-box;
   display: inline-flex;
   position: relative;
   cursor: pointer;
@@ -33,6 +32,8 @@
   background-color: var(--theme-primary);
 
   &:hover {
+    color: var(--color-white);
+    text-decoration: none;
     filter: opacity(80%);
   }
 }
@@ -73,8 +74,10 @@
   font-weight: var(--font-weight-semibold);
   border-radius: 0;
   justify-content: start;
+  min-height: unset;
 
   &:hover {
+    color: var(--theme-primary);
     text-decoration: underline;
   }
 }
--- a/src/Row/index.stories.js
+++ b/src/Row/index.stories.js
@@ -141,7 +141,7 @@ SectionHeaderExample.parameters = {
 
 export const JustifyingContent = () => (
   <div className="nds-typography">
-    <Row justifyContent="end">
+    <Row justifyContent="end" alignItems="center">
       <Row.Item shrink>
         <Button type="plain">Cancel</Button>
       </Row.Item>
--- a/src/base-styles/reset.scss
+++ b/src/base-styles/reset.scss
@@ -1,0 +1,192 @@
+/**
+* reset.css
+* Custom normailze/reset
+*
+* - Normalizes styling across browsers
+* - Sets sensible defaults for base elements
+* - Repeats some key rules from `semantic-ui` that allow us to remove it from consumers
+*/
+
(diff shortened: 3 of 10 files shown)


src/Button/index.scss
```diff
@@ -73,6 +74,7 @@
   font-weight: var(--font-weight-semibold);
   border-radius: 0;
   justify-content: start;
+  min-height: unset;
```
akdetrick, line 77: The deprecated `PlainButton` did not have a min-height and neither should `Button type="plain"`

src/Row/index.stories.js
```diff
@@ -141,7 +141,7 @@ SectionHeaderExample.parameters = {
 
 export const JustifyingContent = () => (
   <div className="nds-typography">
-    <Row justifyContent="end">
+    <Row justifyContent="end" alignItems="center">
```
akdetrick, line 144: now that `<Button type="plain"` no longer has a min height, update code example to vertically align different button types

src/base-styles/typography.scss
```diff
@@ -1,8 +1,4 @@
 .nds-typography {
-  font-family: var(--font-family-default);
-  text-rendering: geometricPrecision;
-  line-height: var(--font-lineHeight-default);
-
```
akdetrick, line 5: These styles are now set globally by `defaults.css`.

src/base-styles/reset.scss
```diff
@@ -0,0 +1,192 @@
...
+* Custom normailze/reset
+*
+* - Normalizes styling across browsers
+* - Sets sensible defaults for base elements
+* - Repeats some key rules from `semantic-ui` that allow us to remove it from consumers
+*/
+
+// border-box on all elements
+html {
+  box-sizing: border-box;
```
wrbrand, line 12: This seemed like the riskiest part of this PR, but looking again this is set in semantic-ui anyway, and our overrides were only to adjust for that in Storybook.

wrbrand (approved):
I've mostly taken a probabilistic approach to reviewing this, but so far all the selectors I've looked at have only had sensible resets here that are accounted for in semantic-ui's reset.

PR #531, opened 2022-01-24
fix(Dialog): address design feedback

### LINKED ISSUE #417: [Dialog] style: design feedback
- [x] remove the `--bordered` style variant from footer; the gradient is enough
- [x] Title color need to be text black, not primary
- [x] Need 8px of spacing after the line before the body text
- [x] "X" needs to be 16px away from the edges on both sides. Currently is more. Also needs to be 12x12 in size and text black
- [x] Padding on all sides should be 40px. Looks like there's a little extra on the top.

----

Screenshot from https://github.com/narmi/banking/issues/14370

akdetrick:
fixes #417 

Updates `Dialog` styling to address design feedback

Screen Shot 2022-01-24 at 3 43 52 PM

### DIFF
--- a/src/Dialog/index.js
+++ b/src/Dialog/index.js
@@ -80,14 +80,17 @@ const Dialog = ({
             <span className="narmi-icon-x"></span>
           </button>
         </div>
-        <div ref={contentRef} className="nds-dialog-content nds-typography">
+        <div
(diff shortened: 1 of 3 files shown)


xiexiewenting (approved):
looked at the storybook component, and it looked great!

PR #537, opened 2022-01-24
Breaking/464 button props

### LINKED ISSUE #464: [Button] (breaking) remove `type`, `children` and `className` props
Make the following breaking changes to `Button`:

- render `button` element by default
- rename `type` to `kind`
- remove `children` (consumers must use `label` moving forward)
- remove `className` (do not allow style customization on buttons, period)

akdetrick:
fixes #464 

**Breaking change merging into `major/v2`**

Making the following changes in individual commits so that release notes display complete information on what breaking changes were made.

- Renders `Button` as a `button` element by default. Consumers may use `as="a"` to render as an anchor element
- ~Removes `children` support. Consumers must use `label` moving forward.~
- Removes `className` support. Consumers must rely on `kind` prop to choose button style from a limited set.
- Renames `type` prop to `kind`

_(Edit: reverted removal of `children` prop for this PR)_

Screen Shot 2022-01-24 at 6 37 39 PM

### DIFF
--- a/src/Button/index.js
+++ b/src/Button/index.js
@@ -16,7 +16,6 @@ const Button = ({
   disabled = false,
   type = "primary",
   label,
-  className,
   onClick = () => {},
   as = "button",
   ...otherProps
@@ -37,7 +36,6 @@ const Button = ({
           resetButton: as === "button",
           "nds-button--disabled": disabled,
         },
-        className,
       ])}
       disabled={isButtonElement && disabled ? true : undefined}
       data-testid="nds-button"
@@ -58,14 +56,6 @@ Button.propTypes = {
   type: PropTypes.oneOf(["primary", "secondary", "menu", "plain"]),
   /** Click callback, with event object passed as argument */
   onClick: PropTypes.func,
-  /**
-   * ️**⚠️ DEPRECATED**
-   *
-   * Support for passing custom `className` strings will be removed in
-   * a future release.
-   * Please use the `type` prop to determine the button style instead.
-   */
-  className: PropTypes.string,
 };
 
 export default Button;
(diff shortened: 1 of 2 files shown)


src/Button/index.js
```diff
@@ -39,47 +31,31 @@ const Button = ({
...
-   * Please use the `type` prop to determine the button style instead.
-   */
-  className: PropTypes.string,
-  /**
-   * **⚠️ DEPRECATED**
-   *
-   * Passing children to render the button label will be removed
-   * in a future release. Use the `label` prop instead.
-   */
-  children: PropTypes.node,
```
graphicsforge, line 82: are you sure you want to do this? I thought we wanted nds usage to be as analogous to normal DOM elements as possible, and `<button>do thing</button>` is [current recommended practice](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#accessibility_concerns)

akdetrick: > nds usage to be as analogous to normal DOM elements as possible

This may have been a stated goal before my time, but my goal is to make NDS components as reusable and composable as possible. The props interface does not need to match the DOM output to accomplish that.

As for the `button` element recommendation, I totally agree. This PR changes the default DOM element rendered to `button` (we were rendering an anchor up until now)

akdetrick: @graphicsforge On closer, inspection it looks like we need to support an `icon` prop before we can do this:

`NewThreadModal.jsx`
```jsx
<Button
          className="ui primary button new-thread-button"
          onClick={(e) => onClick(e)}
        >
          <Plus size={17} />
          New message
        </Button>
```

I will roll back this change for the v2 release and `children` will be supported until we can add icon prop support to `Button`.

graphicsforge (approved):

PR #544, opened 2022-01-26
fix: use ESM export syntax for root export

agreatscott:
Consuming `1.29.0` results in an error when importing the components, this should resolve it, but maybe there's a better way?


akdetrick:
applied the fix to `master`

PR #561, opened 2022-02-03
test(index): basic example consumer unit test

### LINKED ISSUE #560: [test] Add an example consumer / index unit test
Write a simple unit test in `src` root that imports and uses components from `dist/index.js`.

This would catch annoying errors in exports that break _everything_. Those errors are too easy to make, so a test here is very important.

akdetrick:
fixes #560 

A little something to save us from the embarrassment of breaking all exports by making a mistake in `src/index.js`. It's not a full example consumer, but it's better than nothing.

When I recreate the issue we encountered a week or so ago (`module.exports = componentsList`) in the index file, the unit test catches the problem:

Screen Shot 2022-02-03 at 5 55 07 PM

### DIFF
--- a/.github/workflows/jest.yaml
+++ b/.github/workflows/jest.yaml
@@ -19,4 +19,5 @@ jobs:
         run: |
           echo "The job was automatically triggered by a ${{ github.event_name }} event."
           npm install
+          npm run build
           npm run test
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,5 +2,8 @@ module.exports = {
   moduleFileExtensions: ["js", "json", "jsx", "vue"],
   moduleDirectories: ["node_modules", "src"],
   testEnvironment: "jest-environment-jsdom",
-  setupFilesAfterEnv: ["./setupTests.js"],
+  setupFilesAfterEnv: ["<rootDir>/setupTests.js"],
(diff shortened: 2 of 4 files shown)


.github/workflows/jest.yaml
```diff
@@ -19,4 +19,5 @@ jobs:
         run: |
           echo "The job was automatically triggered by a ${{ github.event_name }} event."
           npm install
+          npm run build
```
akdetrick, line 22: the index unit test reads from `dist/`, so we must build before testing now

jest.config.js
```diff
@@ -2,5 +2,8 @@ module.exports = {
   moduleFileExtensions: ["js", "json", "jsx", "vue"],
   moduleDirectories: ["node_modules", "src"],
   testEnvironment: "jest-environment-jsdom",
-  setupFilesAfterEnv: ["./setupTests.js"],
+  setupFilesAfterEnv: ["<rootDir>/setupTests.js"],
+  moduleNameMapper: {
+    "\\.(css|scss)$": "<rootDir>/__mocks__/styleMock.js",
```
akdetrick, line 7: tells jest what to do with imports/requires that are normally handled by webpack

xiexiewenting (approved):
yay!

Issue #344, opened 2021-11-04
[Tag] Add Tag component


akdetrick:
### Short Description
A rounded rectangle inline label. `Tag` should support several variants, including a variant with a dismiss button.

### Mockup

https://zeroheight.com/8ac87d4ba/p/08acdf-tags
<img width="666" alt="Screen Shot 2021-11-04 at 4 56 45 PM" src="https://user-images.githubusercontent.com/231252/140418789-f691e79d-09a2-41b5-81e6-c6a8a95b542f.png">

### Responsibilities

**This component should be responsible for the following:**
- rendering a rounded rect with a label
- firing a callback when the dismiss button is clicked

**This component should NOT be responsible for...**
- doing anything on dismissal click (consumers need to use the callback to update surrounding state)

### Props Signature

```
<Tag
  kind="interactive"
  onDismiss={cbFunc}
  label="my cool tag"
/>
```

```
Tag.propTypes = {
  label: <string>,
  onDismiss: <callback>, // consumers use this to decide what to do with the "x" button click
  kind: oneOf(["interactive", "outline", "solid"]) // only the `interactive` type should have a close button per design
}
```

PR #568, opened 2022-02-11
feat(Tag): Tag component

reshav-abraham:
closes https://github.com/narmi/design_system/issues/344

Initial tag component.

 Referenced spec https://zeroheight.com/8ac87d4ba/p/08acdf-tags


src/Tag/index.scss
```diff
@@ -0,0 +1,30 @@
...
+    color: RGB(var(--rgb-white));
+    cursor: pointer;
+    padding-left: 8px;
+    font-size: 12px;
+    position: relative;
+    top: 1px;
+  }
+}
+.nds-tag--dismissible {
+  color: RGB(var(--rgb-white));
```
akdetrick, line 17: the `--color` variables don't need the RGB wrapper - you can just do:

```css
color: var(--color-white);
```

The `--rgb` variables are for when you need to transparentize a color by mixin in an alpha channel like `rgba(var(--rgb-pine), var(--alpha-20));`

src/Tag/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+ const noop = () => {};
+
+const Tag = ({
+  kind = "subdued", // outline, subdued, x-tag (cactus400) #7fbc5b; #7FBC5B oneof
+  onDismiss = noop,
+  label = "my cool tag",
+}) => {
+  return (
+    // <AsElement
```
akdetrick, line 27: ✂️ we could take out the `AsElement` support for now

src/Tag/index.js
```diff
@@ -0,0 +1,47 @@
+import React, { useState, useEffect } from "react";
+import PropTypes from "prop-types";
+import { useLayer } from "react-laag";
+import cc from "classcat";
+import AsElement from "../util/AsElement";
+
+/**
```
akdetrick, line 7: ✂️ If you update this big jsDoc comment and put it directly above the `const Tag` function, the docs will show up in storybook above the "overview" story

src/Tag/index.js
```diff
@@ -0,0 +1,47 @@
+import React, { useState, useEffect } from "react";
+import PropTypes from "prop-types";
+import { useLayer } from "react-laag";
+import cc from "classcat";
+import AsElement from "../util/AsElement";
```
akdetrick, line 5: don't need `AsElement` for now

src/Tag/index.scss
```diff
@@ -0,0 +1,30 @@
+.nds-tag {
+  padding: 8px;
+  border-radius: 16px;
+  display: flex;
+  align-items: baseline;
+  font-size: 12px;
```
akdetrick, line 6:
```suggestion
  font-size: var(--font-size-xs);
```

src/Tag/index.scss
```diff
@@ -0,0 +1,30 @@
+.nds-tag {
+  padding: 8px;
```
akdetrick, line 2: you could do one of two things here:

use the var:
```css
padding: var(--space-s);
```

or apply the padding via a helper class

```jsx
{/* padding--all--s applies `s` padding (8px) to all sides */}
<div className={`nds-typography nds-tag nds-tag--${kind} padding--all--s`}>
```

akdetrick (changes requested):
Left some comments; looks like this also needs a rebase from master

src/Tag/index.js
reshav-abraham: Wrote a new description. Will defer to any suggestions you have!

src/Tag/context.js
```diff
@@ -0,0 +1,21 @@
+import { createContext } from "react";
```
akdetrick, line 1: Looks like this got copied over by mistake ✂️

src/Tag/index.scss
```diff
@@ -0,0 +1,30 @@
...
+    padding-left: 8px;
+    font-size:var(--font-size-xs);
+    position: relative;
+    top: 1px;
+  }
+}
+.nds-tag--dismissible {
+  color: var(--color-white);
+  background: var(--color-cactus);
+  border: 1px solid var(--rgb-cactus);
```
akdetrick, line 19:
```suggestion
  border: 1px solid var(--color-cactus);
```

akdetrick (changes requested):
Just some minor things, but otherwise looks good

tokens/src/color/narmi.json
```diff
@@ -75,6 +75,7 @@
       "grey": { "value": "#4C4C4C" },
       "mediumGrey": { "value": "#8C8C8C" },
       "lightGrey": { "value": "#D9D9D9" },
+      "smokeGrey": { "value": "#F3F3F3" },
```
akdetrick, line 78: The token `--bgColor-smokeGrey` already exists, but I am now seeing that it's confusing to use a `color` prefix for some tokens and a `bgColor` prefix for others.

I might copy more of those `bgColor` tokens into the `color` category...

akdetrick (changes requested):
a few minor things, but otherwise looks good

akdetrick (approved):

src/Tag/index.js
```diff
@@ -0,0 +1,31 @@
...
+const noop = () => {};
+
+/**
+ * A rounded rectangle inline label.
+ * The user has the option of firing a callback for 'dismissible' Tags.  
+ */
+const Tag = ({
+  kind = "subdued", // outline, subdued, x-tag (cactus400) #7fbc5b; #7FBC5B oneof
+  onDismiss = noop,
+  label = "my cool tag",
```
akdetrick, line 14: might be good to make this an empty string by default - if we're missing data or something else makes the `label` undefined, this message could show up in prod

reshav-abraham: Good call!

tokens/src/color/narmi.json
reshav-abraham: This line actually introduces this warning  while building
```
cssRgbColors
⚠️ /Users/reshavabraham/work/narmi/design_system/dist/tokens/css/rgbColors.css
While building rgbColors.css, token collisions were found; output may be unexpected.
    Output name rgb-smokeGrey was generated by:
        color.background.smokeGrey   243, 243, 243
        color.narmi.smokeGrey   243, 243, 243
This many-to-one issue is usually caused by some combination of:
    * conflicting or similar paths/names in property definitions
    * platform transforms/transformGroups affecting names, especially when removing specificity
    * overly inclusive file filters
```
So I've gone with `--bgColor-smokeGrey`

src/Tag/index.scss
```diff
@@ -0,0 +1,30 @@
+.nds-tag {
+  padding:var(--space-xs);
+  border-radius: 16px;
+  display: flex;
+  align-items: baseline;
+  font-size: var(--font-size-xs);
+  .narmi-icon-x {
+    color: var(--color-white);
+    cursor: pointer;
+    margin-left: 8px;
```
reshav-abraham, line 10: This needs to be margin instead of padding because the padding space makes the clickable area bigger

akdetrick: good catch!

PR #579, opened 2022-02-22
fix(Dialog): close dialog on click outside of content box

agreatscott:
I've noticed that clicking outside of the dialog does not close it - I think expected behavior is that such a click event should in fact dismiss it. 

This PR simply adds another `useEffect` hook with a click listener, to detect if a click event target is outside of the dialog content and if it is, close the dialog.


ethan-hogen (commented):
lgtm, but will defer approval to someone with more js/nds experience

akdetrick (approved):
Looks like I just forgot to do this - thank you for fixing it!

PR #584, opened 2022-02-24
feat(Tabs): add `selectedIndex` prop to enable fully controlled Tabs

### LINKED ISSUE #570: [Tabs] programmatic tab changes not rerendering tabs
### Component/Module
`Tabs`

### Reproduction Steps
Update `defaultSelectedIndex` programmatically

(see: https://github.com/narmi/banking/pull/15819/commits/afad121852f588520d59a3d77f963325112db680)

### Expected Result
Tabs rerenders with the new tab selected

### Actual Result
You must use `key` to force a rerender

This might mean that we need to support a controlled mode for Tabs.

akdetrick:
fixes #570 

Adds `selectedIndex` prop that allows consumers to optionally use `Tabs` as a fully controlled component.

Screen Shot 2022-02-24 at 6 55 02 PM

Screen Shot 2022-02-24 at 6 53 55 PM

### DIFF
--- a/src/Tabs/TabsList.js
+++ b/src/Tabs/TabsList.js
@@ -5,7 +5,7 @@ import TabsContext from "./context";
 const noop = () => {};
 
 const TabsList = ({ children }) => {
-  const { tabIds, setTabIds, changeTabs, selectedIndex, hasPanels } =
+  const { tabIds, setTabIds, changeTabs, currentIndex, hasPanels } =
     useContext(TabsContext);
   const childArray = React.Children.toArray(children);
 
@@ -21,13 +21,13 @@ const TabsList = ({ children }) => {
     let newIndex;
     switch (key) {
       case "ArrowLeft":
-        newIndex = selectedIndex - 1;
+        newIndex = currentIndex - 1;
         if (newIndex >= 0) {
           changeTabs(tabIds[newIndex]);
         }
         break;
       case "ArrowRight":
-        newIndex = selectedIndex + 1;
+        newIndex = currentIndex + 1;
         if (newIndex <= tabIds.length - 1) {
           changeTabs(tabIds[newIndex]);
         }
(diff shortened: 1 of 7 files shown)


src/Tabs/TabsList.js
```diff
@@ -5,7 +5,7 @@ import TabsContext from "./context";
 const noop = () => {};
 
 const TabsList = ({ children }) => {
-  const { tabIds, setTabIds, changeTabs, selectedIndex, hasPanels } =
+  const { tabIds, setTabIds, changeTabs, currentIndex, hasPanels } =
```
akdetrick, line 8: `currentIndex` is the new internal name for the currently selected index. This prevents naming collisions with the new `selectedIndex` prop

src/Tabs/context.js
```diff
@@ -5,9 +5,8 @@ const TabsContext = createContext({
   tabIds: [],
   setTabIds: () => {},
 
-  // index of tab in source order that is selected
-  selectedIndex: 0,
-  setSelectedIndex: () => {},
```
akdetrick, line 10: none of the context consumers needed the setter; better to remove it

src/Tabs/index.js
```diff
@@ -56,10 +62,16 @@ Tabs.propTypes = {
...
    */
   defaultSelectedIndex: PropTypes.number,
-  /** Callback invoked with `tabId` of the tab as the argument */
+  /**
+   * Sets selected tab by index, making Tabs **fully controlled**.
+   * When using this prop, you must use the `onTabChange` callback
+   * to update the value of this prop to update the selected tab.
+   */
+  selectedIndex: PropTypes.number,
+  /** Callback invoked with the index of the tab the user is moving selection to */
```
akdetrick, line 74: The `onTabChange` callback has a different argument now... this might be considered a breaking change, but given the fact that Tabs were uncontrolled previously, this callback was functionally useless until now

graphicsforge (approved):
Yay!
Thanks for this! Most components in banking are fully controlled so this will really help things function as intended!

Issue #572, opened 2022-02-17
[Dropdown] `error` prop unhandled


akdetrick:
### Component/Module
`Dropdown`

### Reproduction Steps
Pass an `error` prop to `Dropdown`

### Expected Result
The dropdown box should be outlined red with an error message rendered.

### Actual Result
The incoming `error` prop is ignored and the Dropdown is rendered normally.

### Additional Details / Context
This behavior works on `DateInput` and `TextInput` because they are composed from `Input`, which handles the incoming `error` prop.
The fix should be replicating the behavior of `Input` in `Dropdown` so it handles an incoming `error` prop.

PR #597, opened 2022-03-03
fix(Dropdown): Allow user to pass error to Dropdown

reshav-abraham:
closes https://github.com/narmi/design_system/issues/572

Pass error prop to TextInput in Dropdown

### DIFF
--- a/src/Dropdown/index.js
+++ b/src/Dropdown/index.js
@@ -122,7 +122,7 @@ Dropdown.propTypes = {
   /** Initial label shown in input */
   triggerLabel: PropTypes.string.isRequired,
   /** Array of elements to render as items in the dropdown */
-  children: PropTypes.arrayOf(PropTypes.element),
+  children: PropTypes.array,
   /** Sets `field` prop of internal `TextInput` */
   field: PropTypes.string,
   /** When true, the dropdown initially renders as open */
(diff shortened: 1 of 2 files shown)


src/Dropdown/index.js
```diff
@@ -121,7 +122,7 @@ Dropdown.propTypes = {
   /** Initial label shown in input */
   triggerLabel: PropTypes.string.isRequired,
   /** Array of elements to render as items in the dropdown */
-  children: PropTypes.arrayOf(PropTypes.element),
```
reshav-abraham, line 124: noticed this error 
checkPropTypes.js:20 Warning: Failed prop type: Invalid prop `children[0]` of type `array` supplied to `Dropdown`, expected a single ReactElement

src/Dropdown/index.js
```diff
@@ -90,6 +90,7 @@ const Dropdown = (props) => {
         field={props.field}
         label={props.triggerLabel}
         readOnly
+        error={props.error}
```
akdetrick, line 93: I'd recommend adding a PropType for the `error` prop; it will show up in the props table and throw helpful errors if someone passes something other than a string

akdetrick (changes requested):
looks good; just need a prop type for `error`

src/Dropdown/index.js
reshav-abraham: 👍

akdetrick (approved):

PR #603, opened 2022-03-07
fix(Tag): make themable; prevent auto height using inline-block display

### LINKED ISSUE #601: [Tag] layout issues
### Component/Module
`Tag`

screen_shot_2022-03-07_at_1 15 04_pm
screen_shot_2022-03-07_at_1 33 57_pm

### Reproduction Steps
Place the `Tag` so that it becomes a flex child - the height and width change

### Expected Result
1. The tag has consistent height and padding, and shrinks to content width, regardless of context.
2. The tag uses theme colors for all variants

Need to replace some `color` values with `theme` values. This should also be an `inline-flex` element that always shrinks to content width.

akdetrick:
fixes #601 

- Changes some colors to use `--theme` colors
- Changes `display` property of the Tag to use `inline-block` so as to avoid auto height/width behavior of a flex
- Uses `Row` to eliminate some extra styles

### DIFF
--- a/src/Tag/index.scss
+++ b/src/Tag/index.scss
@@ -1,30 +1,27 @@
 .nds-tag {
-  padding:var(--space-xs);
+  padding: var(--space-xs) var(--space-s);
   border-radius: 16px;
-  display: flex;
-  align-items: baseline;
+  display: inline-block;
   font-size: var(--font-size-xs);
   .narmi-icon-x {
-    color: var(--color-white);
     cursor: pointer;
-    margin-left: 8px;
-    font-size:var(--font-size-xs);
-    position: relative;
-    top: 1px;
+    font-size: var(--font-size-xs);
+    display: inline-block;
+    transform: translateY(1px);
   }
 }
 .nds-tag--dismissible {
   color: var(--color-white);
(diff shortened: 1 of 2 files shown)


src/Tag/index.scss
```diff
@@ -1,30 +1,27 @@
...
   .narmi-icon-x {
-    color: var(--color-white);
     cursor: pointer;
-    margin-left: 8px;
-    font-size:var(--font-size-xs);
-    position: relative;
-    top: 1px;
+    font-size: var(--font-size-xs);
+    display: inline-block;
+    transform: translateY(1px);
```
akdetrick, line 10: setting display to `inline-block` allows the icon to accept a `transform`

src/Tag/index.scss
```diff
@@ -1,30 +1,27 @@
 .nds-tag {
-  padding:var(--space-xs);
+  padding: var(--space-xs) var(--space-s);
   border-radius: 16px;
-  display: flex;
-  align-items: baseline;
```
akdetrick, line 5: `Row` is providing the centering behavior and the gap between the label and icon

PR #610, opened 2022-03-10
fix(Tag): match tag height to default text height

### LINKED ISSUE #606: [Tag] Height should match line height of default body text
### Component/Module
`Tag`

Screen Shot 2022-03-08 at 12 37 21 PM

We should change y axis padding to `xxs` for starters. The designs call for 10px text, but let's see if we can get the height correct without sacrificing readability.

akdetrick:
fixes #606 

- Matches `Tag` height to default text height
- Adds story with a layout example next to text


xiexiewenting (approved):
tested this in banking on the transactions card, LGTM!

PR #633, opened 2022-03-23
fix(Popover): add focus trap for a11y

### LINKED ISSUE #632: [Popover] base popover should contain a focus trap
### Component/Module
`Popover`

### Expected Result
When a Popover is open, focus should only move within the popover until the user dismisses it.

### Actual Result
Focus can be moved outside the open popover

akdetrick:
fixes #632 

Accessibility enhancement.
When the popover is open, focus is "trapped" in the popover content. This makes it easier to keyboard navigate the contents of a popover.

### DIFF
--- a/src/Popover/index.stories.js
+++ b/src/Popover/index.stories.js
@@ -25,6 +25,23 @@ Overview.argTypes = {
   content: { control: false },
 };
 
+export const FocusManagement = Template.bind({});
+FocusManagement.args = {
+  children: <Button type="secondary">Click to show Popover</Button>,
+  content: (
+    <div className="padding--all">
+      Focus will be trapped to{" "}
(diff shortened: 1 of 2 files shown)


agreatscott (approved):

src/Popover/index.stories.js
```diff
@@ -25,6 +25,23 @@ Overview.argTypes = {
...
+export const FocusManagement = Template.bind({});
+FocusManagement.args = {
+  children: <Button type="secondary">Click to show Popover</Button>,
+  content: (
+    <div className="padding--all">
+      Focus will be trapped to{" "}
+      <a _target="blank" href="http://narmi.com" className="fontWeight--bold">
+        focusable
+      </a>{" "}
+      <a _target="blank" href="http://narmi.com" className="fontWeight--bold">
```
chris-griffin, line 37: Mind as well use https

PR #638, opened 2022-03-24
feat(Select): add `Select` component

### LINKED ISSUE #630: [Select] Deprecate `Dropdown`; add `Select`
It is not possible to make our current `Dropdown` fully accessible due to the current interface of passing arbitrary children. Let's deprecate `Dropdown` and create a separate component that only handles the select ARIA design pattern.

#### `Select.Item`/`Select.Action` components
Add a dedicated child component for Select, `Select.Item` (similar to `Row.Item`). This gives us a couple of things:
- full rendering control of items
- dedicated props for string values of items (for updating the trigger label)
- dedicated props for handling side effects
- no surprises for unsupported element types

```jsx
<Select
  label="Select a name for your horse"
  onChange={onHorseNameChange}
>
  <Select.Item value="foo">Foo item</Dropdown.Item>
  <Select.Item value="bar">Bar item</Dropdown.Item>
  <Select.Action onClick={() => { alert('side effects happen here') }}>
    Any JSX can be rendered here (good for icons + text or anything else)
  </Select.Item>
</Select>
```

akdetrick:
fixes #630 

Add `Select` component and deprecates `Dropdown`. We are going the route toward dropdowns tailored to their specific needs rather than a general purpose component which can not easily be made accessible.

This uses our `DropdownTrigger` in combination with the `downshift` library for managing a11y attributes, keyboard events, and focus.

----

## 🔍 Reviewing this PR

I recommend getting context for these changes by running the branch locally:

```sh
git checkout -t origin/feat/630-select-component
npm install
npm run build
npm run storybook
```

## Example
Mar-24-2022 16-01-21

### DIFF
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -1,0 +1,156 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { useSelect } from "downshift";
+import cc from "classcat";
+import DropdownTrigger from "../DropdownTrigger";
+
+const noop = () => {};
+
+/**
+ * @param {Object} item a Select.Item or Select.Action component
+ * @returns {Boolean} true if the item is a Select.Action
+ */
+const isAction = (item) => {
+  let result = false;
+  if (item) {
+    result = item.type.name === "SelectAction";
+  }
+  return result;
+};
+
+/**
+ * Sets the selected value in the DropdownTrigger.
+ * A Select.Action should _not_ display as selected in the button.
+ *
+ * @param {Object} item the currently selected Select.Item or Select.Action
+ * @returns {String|Node} The value to display in the trigger button
+ */
+const getSelectedItemDsiplay = (item) => {
+  let result = "";
+  if (item && !isAction(item)) {
+    result = item.props.children;
+  }
+  return result;
+};
+
+/**
+ * Accessible custom select control for giving users the ability to select one option from a list of options.
+ * `Select` supports the ability to pass in a `<Select.Action>` that acts as an option that only triggers a side effect.
+ */
+const Select = ({
+  label,
+  children,
+  onChange = noop,
+  defaultValue,
+  defaultOpen = false,
+}) => {
+  // only include valid Select children in options items
+  const items = React.Children.toArray(children).filter((child) =>
+    ["SelectItem", "SelectAction"].includes(child.type.name)
+  );
+
+  const defaultSelectedItem = items
+    .filter((item) => !isAction(item))
+    .filter((item) => item.props.value === defaultValue)
+    .pop();
+
+  const {
+    isOpen,
+    selectedItem,
+    getToggleButtonProps,
+    getLabelProps,
+    getMenuProps,
+    highlightedIndex,
+    getItemProps,
+  } = useSelect({
+    items,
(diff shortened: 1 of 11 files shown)


src/Select/index.test.js
```diff
@@ -0,0 +1,109 @@
...
+import { render, screen, fireEvent, prettyDOM } from "@testing-library/react";
+import Select, { isAction, getSelectedItemDsiplay } from "./";
+
+describe("Select", () => {
+  /**
+   * Most of the complexity of this component is in `DropdownTrigger` and
+   * `useSelect` from downshift, which have their own test coverage.
+   */
+
+  it("isAction: detects action items corrlectly", () => {
```
ejrb, line 11:
```suggestion
  it("isAction: detects action items correctly", () => {
```

src/Select/index.js
```diff
@@ -0,0 +1,149 @@
...
+  defaultOpen: PropTypes.bool,
+  /**
+   * Error message.
+   * When passed, this will cause the trigger to render in error state.
+   */
+  errorText: PropTypes.string,
+};
+
+Select.Item = SelectItem;
+Select.Action = SelectAction;
```
ejrb, line 148: neat! 📸

src/Select/index.js
```diff
@@ -0,0 +1,149 @@
...
+};
+
+/**
+ * Sets the selected value in the DropdownTrigger.
+ * A Select.Action should _not_ display as selected in the button.
+ *
+ * @param {Object} item the currently selected Select.Item or Select.Action
+ * @returns {String|Node} The value to display in the trigger button
+ */
+export const getSelectedItemDsiplay = (item) => {
```
ejrb, line 32:
```suggestion
export const getSelectedItemDisplay = (item) => {
```

src/Select/index.js
```diff
@@ -0,0 +1,149 @@
...
+  /** Label for the select control */
+  label: PropTypes.string.isRequired,
+  /** Change callback. Called with value string from the selected item */
+  onChange: PropTypes.func,
+  /**
+   * Use to set a default selection by passing the `value` prop
+   * of one of the `<Select.Item>` children.
+   */
+  defaultValue: PropTypes.string,
+  /** Open the dropdown on rendder if `true` */
```
ejrb, line 138:
```suggestion
  /** Open the dropdown on render if `true` */
```

src/Select/index.scss
```diff
@@ -0,0 +1,30 @@
+// make sure we don't need to flex center the outer wrapper??
```
ejrb, line 1: well, do we?

ejrb (commented):
Really nice - couple of small typos.  Thanks for the download instructions in the PR description!

src/Select/index.scss
akdetrick: apparently not! removing the comment

src/Select/index.js
akdetrick: we do the same thing in `Tabs` so that `import { Tabs } from '@narmi/design_system';` pulls in the main component and its subcomponents like `Tabs.Panel`.

ejrb (approved):

PR #639, opened 2022-03-24
docs(storybook): add props tables for subcomponents

akdetrick:
Configure stories to output props tables for sub components like `Row.Item` and `Tabs.Panel`.

Screen Shot 2022-03-24 at 2 50 29 PM
Screen Shot 2022-03-24 at 2 56 21 PM

### DIFF
--- a/src/Row/RowItem.js
+++ b/src/Row/RowItem.js
@@ -1,0 +1,29 @@
+import React from "react";
+import PropTypes from "prop-types";
+import cc from "classcat";
+import AsElement from "../util/AsElement";
+
+/**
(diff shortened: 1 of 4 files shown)


src/Row/RowItem.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
```
akdetrick, line 1: This probably should have been broken out into its own file in the first place, but was necessary for this documentation feature because react-docgen associates _all_ proptypes it finds in a file with the default exported component.

adamcadaver (approved):

PR #648, opened 2022-03-30
fix(RadioButtons): fix keyboard navigation for a11y

### LINKED ISSUE #634: [RadioButtons] Make options keyboard navigable
### Component/Module
`RadioButtons`

### Expected Result
RadioButtons are keyboard navigable.

### Actual Result
RadioButton elements are not focusable at all.
The arrow keys of course won't work this way.

akdetrick:
fixes #634 

When you use `display: none`, the element is fully removed from document flow and the browser treats the element as if it is not in the DOM. Screen readers and keyboard navigation behaves as if the element doesn't exist.

We were hiding the native `input type="radio"` elements using `display: none`. This PR uses some different CSS to visually hide the input without making the element inaccessible to screen readers and keyboard navigation.

### DIFF
--- a/src/RadioButtons/index.scss
+++ b/src/RadioButtons/index.scss
@@ -7,21 +7,24 @@
     padding-left: 34px;
     margin-bottom: var(--space-m);
     cursor: pointer;
-    font-size: 16px;
     line-height: 1.25;
-    font-family: var(--nds-font-family);
     font-weight: 400;
 
     input {
       position: absolute;
-      opacity: 0;
-      display: none;
+      clip: rect(0 0 0 0);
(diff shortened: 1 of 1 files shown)


src/RadioButtons/index.scss
```diff
@@ -7,21 +7,24 @@
     padding-left: 34px;
     margin-bottom: var(--space-m);
     cursor: pointer;
-    font-size: 16px;
     line-height: 1.25;
-    font-family: var(--nds-font-family);
```
akdetrick, line 12: not needed - NDS base styles provide this now

ejrb (approved):
arrow keys work for me 👍

PR #649, opened 2022-03-30
fix(Checkbox): enable keyboard navigation for a11y

### LINKED ISSUE #642: [Checkbox] must be keyboard navigable
### Component/Module
`Checkbox`

### Expected Result
Label or input of checkbox must be focusable

akdetrick:
fixes #642 

Similar to #648 - when an element is fully hidden with `display: none` _or_ has `0, 0` dimensions, the browser and screen readers treat the element like it doesn't exist, which is why we were missing keyboard navigability on our checkbox component.

The fix is to hide the native input with a CSS technique that only hides it visually.

### DIFF
--- a/src/Checkbox/index.scss
+++ b/src/Checkbox/index.scss
@@ -11,16 +11,23 @@
   align-items: center;
   input {
     position: absolute;
-    opacity: 0;
     cursor: pointer;
-    height: 0;
-    width: 0;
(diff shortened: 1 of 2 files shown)


akdetrick:
added story that should demonstrate tabbing through multiple checkboxes:

Screen Shot 2022-03-31 at 4 00 08 PM

PR #656, opened 2022-04-03
chore(formatters): minor correction to jsdoc for formatDate function

agreatscott:
Currently the documentation looks as below, which is not quite correct. For `short` type formatting, years are outputted as two digits. 

image


akdetrick (approved):
thank you for fixing this!

Issue #663, opened 2022-04-06
[Radiobuttons] add unit tests


akdetrick:
It looks like we don't have any test coverage for `Radiobuttons`. We should write some tests. See `Checkbox` for guidance.

- Arrow keys usable to move selection
- Focusable
- Only one in a set should be selected at a time
- Click only selects, never deselects

PR #664, opened 2022-04-06
Feat/616 radiobuttons card variant

### LINKED ISSUE #616: [Checkbox,RadioButtons] Card variant
### Short Description

A single line, interactive card that can be toggled on or off, like a checkbox OR toggled between like a radio group.

### Mockup
Screen Shot 2022-03-16 at 4 10 38 PM

### Implementation

These should be variants via `kind="card"` in `RadioButtons` and `Checkbox` in order to take advantage of the underlying form elements and built-in accessibility.

akdetrick:
fixes #616 

- Added support for "card" variant with `kind` prop
- Added internal state to manage focus/checked styling to ensure a11y for both variants
- Added missing prop definition for `onClick`

This component seems to be missing unit tests - we'll add those in a [separate ticket](https://github.com/narmi/design_system/issues/663)

Screen Shot 2022-04-06 at 10 22 23 AM

### DIFF
--- a/src/RadioButtons/index.js
+++ b/src/RadioButtons/index.js
@@ -1,47 +1,94 @@
-import React, { useState, useEffect } from "react";
+import React, { useState } from "react";
 import PropTypes from "prop-types";
+import cc from "classcat";
 
 /*
 The Narmi RadioButtons component expects an "options" Prop, which is an object where the keys are the radiobutton
 labels and the values are the radiobutton values. An "initialvalue" Prop can be passed to set a default checked
 radiobutton.
 
-i.e.
-    <RadioButtons
-      initialValue={"true"}
-      options={{ "True": "true", "False": "false" }}
-    />
+```
+  <RadioButtons
+    initialValue={"secondValue"}
+    options={{ "First Label": "firstValue", "Second Label": "secondValue" }}
+  />
+```
 */
 
-const RadioButtons = ({ options, name, initialValue, ...containerProps }) => {
+const RadioButtons = ({
+  options,
+  name,
+  initialValue,
+  kind = "normal",
+  onChange = () => {},
+  ...containerProps
+}) => {
+  const [checkedValue, setCheckedValue] = useState(initialValue);
+  const [focusedValue, setFocusedValue] = useState(null);
+
+  const handleChange = (e) => {
+    setCheckedValue(e.target.value);
+    onChange(e);
+  };
+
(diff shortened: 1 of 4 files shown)


src/RadioButtons/index.js
```diff
@@ -1,47 +1,101 @@
-import React, { useState, useEffect } from "react";
+import React, { useState } from "react";
 import PropTypes from "prop-types";
+import cc from "classcat";
 
 /*
```
akdetrick, line 5:
```suggestion
/**
```

src/RadioButtons/index.js
```diff
@@ -1,47 +1,101 @@
...
-      initialValue={"true"}
-      options={{ "True": "true", "False": "false" }}
-    />
+```
+  <RadioButtons
+    initialValue={"secondValue"}
+    options={{ "First Label": "firstValue", "Second Label": "secondValue" }}
+  />
+```
 */
```
akdetrick, line 17:
```suggestion
```

src/RadioButtons/index.js
```diff
@@ -1,47 +1,100 @@
...
+          <input
+            type="radio"
+            checked={checkedValue === value}
+            onChange={handleChange}
+            onFocus={handleFocus}
+            onBlur={handleBlur}
+            value={value}
+            name={name}
+          />
+          <div role="img" className="nds-radio" />
```
akdetrick, line 69:
```suggestion
```

ejrb (commented):
This might be nothing, but I noticed that two of the groups of radio buttons in the story are connected by arrow key movement.  

I'm just pressing the down key in this GIF:
radio-buttons

What I found odd is that it ignores the favorite colors radios.

src/RadioButtons/index.stories.js
```diff
@@ -36,7 +36,29 @@ export const Example = () => {
...
 };
 
+export const AsCard = Template.bind({});
+AsCard.args = {
+  options: {
+    OptionA: "A",
+    OptionB: "B",
+    OptionC: "C",
+  },
+  name: "options",
```
akdetrick, line 46:
```suggestion
  name: "card_options",
```

akdetrick:
@ejrb Good catch - this should be fixed now. The issue was that the cards story shared the same `name` value as the overview story

ejrb (approved):
makes sense, approved 👍

PR #665, opened 2022-04-06
feat(Select): add `value` prop to enable fully controlled mode

### LINKED ISSUE #580: [Dropdown] Allow component to be a Controlled Component
As it stands, the inner value for the Dropdown components is inaccessible because the component is uncontrolled. There are use cases when a developer/user wants to clear the selection from the Dropdown and have the component essentially "reset" both its value and its label, however this use case is impossible to achieve without wiring up the inner value to State (i.e. converting this component into a controlled component).

Suggestion: Expose the `value` prop from the inner `TextInput` component

akdetrick:
fixes #580 

Adds `value` prop to `Select` to enable a fully controlled mode.

Screen Shot 2022-04-06 at 12 42 08 PM

### DIFF
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -42,7 +42,7 @@ export const getSelectedItemDisplay = (item) => {
  */
 export const getItemByValue = (value, items) => {
   const founditem = items
-    .filter((item) => !isAction(item)) // action items may not be selected by default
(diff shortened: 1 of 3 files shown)


graphicsforge (approved):
Thank you!!
All narmi uses of our input widgets will be fully controlled, so this was needed!

PR #666, opened 2022-04-07
fix(Select): minor style fixes; show example in Dialog

### LINKED ISSUE #641: [Select] (follow up) use a portal or `react-laag` for use in Dialog
Right now the dropdown in `Select` renders in-place in the DOM. This won't work for a Select inside of a `Dialog` because the Dialog will cut it off.

We need to portal the dropdown to document body and adopt a new positioning strategy. Options:

1. `react-laag` hooks might be useful here
2. getClientBoundingRect + fixed position + ReacDOM.createPortal

akdetrick:
fixes #641 

Added some minor style tweaks and story example for a `Select` inside a `Dialog`.  

Basically, if there's room, it will just open normally. If there's not enough room under the `Select` for its dropdown, the parent container will grow to fit it.

This isn't my favorite option visually, but it did turn out to be the best option to preserve accessibility of the focus trap in the Dialog.

https://user-images.githubusercontent.com/231252/162219871-36c1c1dd-edfb-451e-8ebf-1cb70d55acad.mp4

### DIFF
--- a/src/Dialog/index.scss
+++ b/src/Dialog/index.scss
@@ -18,7 +18,6 @@
   border-radius: var(--border-radius-default);
   background-color: var(--bgColor-white);
   box-shadow: var(--elevation-high);
-  overflow: hidden;
   display: flex;
   flex-direction: column;
   flex-wrap: nowrap;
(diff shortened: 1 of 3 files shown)


henderson2350 (approved):
lgtm!

Issue #482, opened 2021-12-10
[Tests,Build,CI] Introduce eslint; enable all rules


akdetrick:
- ensure prettier is enabled
- enable linting with defaults for react and a11y
- enforce rules and naming conventions outlined in contributing guidelines

---

### Remaining work

Enable the following lint options one PR at a time, fixing the errors:

- [ ] `jsx-a11y/click-events-have-key-events`
- [ ] `jsx-a11y/no-static-element-interactions`
- [ ] `jsx-a11y/no-noninteractive-tabindex`
- [x] `jsx-a11y/anchor-is-valid`
- [x] `jsx-a11y/label-has-associated-control`
- [x] `jsx-a11y/interactive-supports-focus`
- [x] `react/prop-types`
- [x] `react/jsx-key`
- [x] `no-unused-vars`

Add plugins:

- [ ] [boolean prop name](https://github.com/narmi/design_system/blob/master/contributing/naming-conventions.md#props) validation

PR #671, opened 2022-04-08
build(lint): introduce eslint and lint workflow for PRs

akdetrick:
relates to:
- #482 

Introduces eslint with correct configuration. I only fixed a handful of errors and disabled most of the offending rules so we can address them one by one in follow up PRs.

 This PR will also cause future PRs to run a lint check via github actions.


.eslintrc.json
```diff
@@ -1,12 +0,0 @@
-{
```
akdetrick, line 1: this file was a dead stub

.eslintignore
```diff
@@ -0,0 +1,19 @@
+# deps and build artifacts
+node_modules/
+dist/
+storybook-static/
+selection.json
+
+# dirs without any JS source
+src/base-styles/
+
+# ignored file types
```
akdetrick, line 10: It might seem weird to ignore a bunch of file types instead of specifying `--ext` in the eslint config, but this is intentional.

When adding a new file type to `src`, it forces a decision on whether or not to ignore the extension or add the appropriate eslint plugin to handle the file.

The `ext` option is less likely to be updated when new file types are introduced to the linted dirs.

PR #681, opened 2022-04-11
Feat/662 responsive dialog

### LINKED ISSUE #662: [Dialog] support mobile drawer mode
For mobile size viewports, `Dialog` should take on a "drawer" presentation. Ideally this can be done via CSS only without introducing or changing any props of `Dialog`.

## Mockup

This "make a payment" feature is displayed on desktop in a `Dialog`. For mobile web sizes, this should be a "takeover" modal.

Screen Shot 2022-04-05 at 5 29 06 PM

akdetrick:
fixes #662 

- Sets up new mobile first breakpoints and sass helpers
- Update `Dialog` styles to take over the full screen by default and act as a floating box for "s" or larger viewports

https://user-images.githubusercontent.com/231252/162836869-03f10a1a-a65c-49af-94bc-44710aa362b6.mp4

### DIFF
--- a/src/base-styles/grid.scss
+++ b/src/base-styles/grid.scss
@@ -1,5 +1,10 @@
 $column_count: 12;
 
+$mobile_size: map.get($breakpoints, "s");
+$tablet: map.get($breakpoints, "m");
+$desktop-small: map.get($breakpoints, "l");
+$desktop-big: map.get($breakpoints, "xl");
+
 .nds-container {
   display: block;
   margin: 0 auto;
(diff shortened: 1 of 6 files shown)


src/base-styles/grid.scss
```diff
@@ -1,5 +1,10 @@
 $column_count: 12;
 
+$mobile_size: 375px; // legacy breakpoint size for mobile
+$tablet: map.get($breakpoints, "m");
+$desktop-small: map.get($breakpoints, "l");
+$desktop-big: map.get($breakpoints, "xl");
```
akdetrick, line 6: `grid.scss` makes extensive use of breakpoints that are not all mobile first, so this maps values from the new breakpoints map to the legacy sass vars

aundus (approved):

PR #707, opened 2022-05-02
fix(RadioButtons/Checkbox): explicitly set font size

akdetrick:
related to: https://github.com/orgs/narmi/projects/5/views/13

The `RadioButtons` and `Checkbox` components should just inherit the default font size, but we're finding we must set them explicitly to work around legacy CSS in `banking` overriding the property


graphicsforge (approved):
Oh! I guess you were expecting font rules to have been set up logically, but that wasn't the case in banking unfortunately! :neutral_face: 

Thank you so much!! :heart:

PR #708, opened 2022-05-02
fix(numberformat): Add native JS failover

aundus:
Adding further browser support.
For https://github.com/narmi/banking/issues/18403


akdetrick (approved):

akdetrick:
Alternatively we could just return the input string unformatted for unsupported browsers

PR #711, opened 2022-05-07
fix(TextInput): enable horizontal scrolling for long unbroken lines

akdetrick:
fixes https://github.com/narmi/banking/issues/17201

Prevent long unbroken lines of text in a textarea from exceeding the bounds of its parent box.

### DIFF
--- a/src/Input/index.scss
+++ b/src/Input/index.scss
@@ -168,6 +168,7 @@
   &,
   & > textarea {
     max-height: 80vh; // avoid growing larger than viewport height
+    overflow-wrap: anywhere; // always break line if it exceeds available width
   }
 
   // This pseudo-element is an auto height invisible container that


src/Input/index.scss
```diff
@@ -28,6 +28,10 @@
     display: flex;
     align-items: center;
 
+    // long, unbroken lines of text should horizontal scrol if they exceed
```
wrbrand, line 31: "scroll"

wrbrand (approved):
LGTM!

src/Input/index.scss
```diff
@@ -28,6 +28,10 @@
...
+    // the available width
+    overflow-x: scroll;
```
aundus, line 33: Is there any way that we can make this wrap? I think scrolling still makes it difficult to edit something like certificates.

akdetrick: That's a good point - we could `break-word` instead. This won't respect line breaks pasted in, but it looks like in `banking` we don't respect line breaks anyway and render the `\n` escapes in the field value

aundus (approved):
Thanks!

PR #715, opened 2022-05-10
fix(ContentCard): update styles to match design specs

### LINKED ISSUE #705: [ContentCard] design feedback
- [ ] remove weird checkmark in example story
- [ ] cards should have an actual background (not transparent) by default

akdetrick:
fixes #705 

- make default `ContentCard` background white instead of transparent
- clean up story for interactive `ContentCard`


graphicsforge (approved):
Oh yeah, they're always white, might as well make them white here!

PR #716, opened 2022-05-10
fix(Card fields): match design spec for card variants of radios and c…

### LINKED ISSUE #704: [RadioButtons,Checkbox] design feedback
For `kind="card"` variants of both `Checkbox` and `RadioButtons` components:

- [ ] hover state should not have a background (border change only)
- [ ] labels should _always_ be `theme-primary`

akdetrick:
fixes #704 

Updates the `kind="card"` variants of `Checkbox` and `RadioButtons`.

Basically, the subdued primary background should only appear when the card is in its selected state. Hover states should only have a primary border (no background).

### DIFF
--- a/src/Checkbox/index.scss
+++ b/src/Checkbox/index.scss
@@ -74,14 +74,13 @@
     display: none;
   }
 
+  &:hover,
   &.nds-checkbox--checked,
   &.nds-checkbox--focused {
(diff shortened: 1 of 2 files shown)


DM-SrishtiLulla (approved):
looks great to me!

PR #726, opened 2022-05-16
feat(helper-classes): add xxl spacing size

agreatscott:
Design specs sometimes ask for a spacing of `60px` which for now we hardcode or define as a var in the NDS consumer. It's come up often enough for me to think that it should be made part of the standard.


akdetrick (approved):
Thanks!

PR #737, opened 2022-05-27
fix(formatting): Expect decimal representation in backfall percent formatter

aundus:
For https://github.com/narmi/banking/issues/19538


elliot-wilson (approved):
Classic!

PR #738, opened 2022-05-31
feat: add  prop to all components

### LINKED ISSUE #709: [feat] add `testId` prop to every component
To aid the creation of webdriver style tests in consumers, add a `testId` prop to every NDS component:

```
testId: PropTypes.string,
```

The string should be rendered as the value of a a`data-testid` attribute. The attribute generally belongs of the root element of a given component, but may vary depending on the component. The element receiving the test id should be noted in the jsDoc above the prop type definition so it is clear in public documentation.

akdetrick:
fixes #709 

Adds dedicated `testId` prop to all components.

Not all components spread props, so an explicit, documented prop is used for the sake of API consistency across all components.

Older hard-coded test ids will still be present unless the consumer passes a `testId` override.

### DIFF
--- a/src/Button/index.js
+++ b/src/Button/index.js
@@ -23,6 +23,7 @@ const Button = ({
   kind = "primary",
   startIcon = null,
   endIcon = null,
+  testId,
   children,
   label,
   onClick = () => {},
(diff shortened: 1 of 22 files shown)


ethan-hogen (approved):

agreatscott (approved):
All seem to be added to sensible elements, nice!

PR #752, opened 2022-06-13
Feat/676 combobox

### LINKED ISSUE #676: [Combobox] new combobox component
### Mockup

I feel like there are a couple of these in the figmas so I'm not sure which one we want to initially adopt, but I think any combobox is probably fine and we can iterate on it until we come up with a design that we are comfortable applying everywhere?

### Responsibilities
not really well defined well yet tbh, but should probably wrap both dropdown selector and controlled text input components. Unclear if we need to do a server call (technically needed for tags, but those suck and the api call would take seconds to do) on debounced text input, would probably need to eventually get an optional api query function passed in but might be a puntable feature

akdetrick:
fixes #676 

- Add support for `endIcon` to `TextInput` (used for dropdown indicator chevron in Combobox)
- Enable ref forwarding in `TextInput` (needed for Combobox's `getInputProps()`, which applies a ref)
- Fix `displayName` of subcomponents (e.g. `Select.Action`)
- Add `Combobox` component, stories, and tests

Screen Shot 2022-06-13 at 11 42 53 AM
Screen Shot 2022-06-13 at 11 43 01 AM
Screen Shot 2022-06-13 at 11 43 30 AM


src/Combobox/index.js
```diff
@@ -0,0 +1,235 @@
...
+    return (
+      <TextInput
+        error={errorText}
+        label={label}
+        value={inputValue}
+        startIcon={icon}
+        onChange={onInputChange}
+      />
+    );
+  }
```
graphicsforge, line 61: Hmm, how would you feel about going through the full downshift stuff even for empty arrays? I don't think inputValue is defined just yet, and we'll probably want to still be firing onInputValueChanged() in case this is something that's doing some debounced server call to load in options based on what the user has started to type? :thinking:

graphicsforge (commented):
in `Combobox.propTypes` would it be possible to pass in an initial value? (i.e. so that I can say that a certain industry was selected initially when someone goes back to edit their org from the review page)

graphicsforge (approved):
This looks good to me!
I had a few small additional feature requests but those can easily be done in another ticket even if you potentially agree to them

src/Combobox/index.js
akdetrick: Good point - I think I need to pass something different to `value` here either way

akdetrick: Fixed - and added a story to clarify the behavior

Screen Shot 2022-06-13 at 6 48 49 PM

akdetrick:
@graphicsforge Thanks for the review! I held off on implementing default selection in favor of waiting until someone actually needed it, but it sounds like we probably do need it for Combobox. I added this follow up ticket to add it: https://github.com/narmi/design_system/issues/754

PR #757, opened 2022-06-21
refactor(tokens): replace all legacy `--nds-` tokens with standard vars

### LINKED ISSUE #749: [tokens] Replace all legacy theme variables with design token properties
The following legacy vars may not work in third party consumers of the design system that do not have the shims that `banking` provides. Let's replace them with the appropriate CSS custom property from design tokens:

- [x] `RGB(var(--nds-lightest-grey))`
- [x] `RGB(var(--nds-white))`
- [x] `RGB(var(--nds-dropshadow-light))`
- [x] `RGB(var(--nds-medium-grey))`
- [x] `RGB(var(--nds-red))`
- [x] `RGB(var(--nds-smoke-grey))`
- [x] `RGB(var(--nds-grey))`
- [x] `RGB(var(--nds-black))`

akdetrick:
fixes #749 

Replaces all usage of older `--nds-` CSS vars with standard vars from design tokens. This ensures nothing in NDS relies on the consumer to shim these properties.

### DIFF
--- a/src/Card/index.scss
+++ b/src/Card/index.scss
@@ -1,12 +1,11 @@
 .nds-card {
   width: 100%;
   height: 100%;
-  box-shadow: RGB(var(--nds-dropshadow-light));
   border-radius: 4px;
 
   padding: 20px;
-  background-color: RGB(var(--nds-white));
-  border: 1px solid RGB(var(--nds-lightest-grey));
+  background-color: var(--color-white);
+  border: 1px solid var(--color-lightGrey);
 
   &[data-hoverable="true"] {
     box-shadow: none;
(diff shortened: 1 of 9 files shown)


src/Card/index.scss
```diff
@@ -1,12 +1,11 @@
 .nds-card {
   width: 100%;
   height: 100%;
-  box-shadow: RGB(var(--nds-dropshadow-light));
```
xiexiewenting, line 4: do we need to add back the `box-shadow` using a standard var?

akdetrick: I should have mentioned this - this var has been undefined for at least 8 months. The border defined in a declaration below this handles the card borders, so we can just remove this one

PR #765, opened 2022-06-30
fix(icons): add icon aliases for mobile support

akdetrick:
Adds aliases to handle FIs that have older icon names in settings for the mobile app.

(related: https://gist.github.com/akdetrick/c07d546df094f0716a2b5216281e4149)


wrbrand (approved):
Sweet!

PR #785, opened 2022-07-20
feat(Combobox): add `inputValue` prop to enable fully controlled input

### LINKED ISSUE #778: [Combobox] no `value` prop for fully controlling input
### Component/Module
`Combobox`

Report from Ethan:
> [Grasshopper is] using the Combobox component for our users to select a country. The components are controlled, and the value selected is being persisted with state. When the component re-renders, it appears blank, likely because there is no native "value" field for the component itself. Looking at the code, they noticed it's not actually an HTML input. Could this, or something functionally similar, be added? They're saying it's confusing to users when they need to make an edit and the country field appears blank even though there is an actual value being persisted in state.

Looks like we need to allow the underlying `input` to be fully controlled, or at least accept a `defaultValue`.

### Development plan

- [ ] Create a "Fully Controlled" story for `Combobox`
- [ ] Add a `value` prop that allows parent state to control the value of the input
- [ ] Ensure that user selection can override the input value via the `onChange` callback

akdetrick:
fixes #778 

Allows the input inside `Combobox` to be fully controlled. Addresses an issue Grasshopper found where the input was blank after rerendering in their application.

Screen Shot 2022-07-20 at 4 39 31 PM

### DIFF
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -39,6 +39,7 @@ const Combobox = ({
   label,
   onChange = noop,
   onInputChange = noop,
+  inputValue: controlledInputValue,
   children,
   disableFiltering = false,
   errorText,
@@ -70,6 +71,7 @@ const Combobox = ({
     reset,
   } = useCombobox({
     items: displayedItems,
+    inputValue: controlledInputValue,
     itemToString: (item) => item.props.value,
     onInputValueChange: ({ inputValue }) => {
       // Typeahead behavior - we adjust the list of available options passed
@@ -98,10 +100,9 @@ const Combobox = ({
         newSelection = selectedItem.props.value;
       }
       onChange(newSelection);
+      onInputChange(newSelection);
     },
   });
-
-  const hasSelectedItem = !!selectedItem;
 
   // It is possible that a consumer may have nothing to pass to `children`.
   // For example, if an API response hasn't completed to load in the autocomplete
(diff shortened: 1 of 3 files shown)


src/Combobox/index.js
```diff
@@ -214,6 +218,12 @@ Combobox.propTypes = {
   label: PropTypes.string.isRequired,
   /** Change callback. Called when an item is selected, with the `value` of the selected item */
   onChange: PropTypes.func,
+  /**
+   * Sets value of the input in a controlled manner.
+   * When using the `value` prop, you **must** update it via the
```
akdetrick, line 223:
```suggestion
   * When using the `inputValue` prop, you **must** update it via the
```

src/Combobox/index.js
```diff
@@ -113,10 +114,13 @@ const Combobox = ({
         label={label}
         startIcon={icon}
         onChange={onInputChange}
+        value={inputValue}
       />
     );
   }
 
+  const hasSelectedItem = !!selectedItem;
```
DM-SrishtiLulla, line 122: what does the double exclamation do?

graphicsforge: It's a cast to boolean! Usually not needed for vanilla JS, and more handy in typed languages!

graphicsforge (approved):
👍

src/Combobox/index.js
xiexiewenting: it converts it to the truthy value, basically

PR #792, opened 2022-07-27
fix(icons): make all glyphs equal width; pixel hint and center all shapes

### LINKED ISSUE #760: [icons] Clean up icomoon set; improve update process
There are some things we can do to would improve our set of icons on web and mobile.

### Motivation

The set in icomoon is not standardized on the same grid. Not all icons are scaled or centered consistently, are missing pixel hinting, and are not using the same baseline value as our body font. We spend a lot of time in design review and development tweaking position and scale of icons with a slow feedback loop.

The update process is manual and prone to error. If we make a mistake copying and updating assets, we could either break NextJS consumers or bloat the CSS size of design system to an unacceptable size.

### Changes in icomoon

- Set all icons on a `24` grid (except venmo logo)
- Pixel hint all icons
- Center and scale icons consistently
- In font metadata, set baseline height to match Mulish
- Version the icons in icomoon metadata (bump the major)

All of these settings will persist in the `selection.json`.

## Changes in design_system

- Add README to `src/icons` 
- ~Write a script to automate the update process given that source and destination files are different and the process is prone to error.~

Must create a specimen to show design for approval before merging.
Must bump the minor version of NDS - this is more notable than a fix.
Must include a separate `banking` ticket to remove legacy style hacks as the new minor version of NDS is introduced.

akdetrick:
fixes #760 

Changes applied via icomoon:
- Use a 24 grid for all icons, except when 32 is needed for pixel hinting
- Use a square canvas for all icons (except for `venmo`)
- Center icons within their canvas
- Pixel hint icons

**No icons have been added or removed; only updated**


akdetrick:
Tested this against `azul` in a beta release; seems to be fine.

agreatscott (approved):
thank you 🙏  consistent icons will be super helpful

PR #794, opened 2022-08-05
fix(Select): cover deselection case in controlled mode

### LINKED ISSUE #793: Select controlled value prop undefined throws error
### Component/Module
Select
https://narmi.github.io/design_system/?path=/story/components-select--controlled
Screen Shot 2022-08-04 at 12 18 25

### Reproduction Steps
1. use `Select`
2. set value=undefined or empty string `''`
or 
1. go to https://narmi.github.io/design_system/?path=/story/components-select--controlled
2. click dropdown
3. screenshot above will be thrown
### Expected Result
initial prop value should just be empty
### Actual Result
throws error

akdetrick:
fixes #793 

Fixes errors thrown when `value` is passed, but with a value of `undefined`, `null`, or `""`.

Basically, the component uses `value` to look for a matching `Select.Item` by its `value` prop. If the item is not found in children, the Selection should clear. There were a few places where the code assumed an item would always be present and have props.

### DIFF
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -14,7 +14,7 @@ const noop = () => {};
  */
 export const isAction = (item) => {
   let result = false;
-  if (item) {
+  if (item && item.props) {
     result = "onSelect" in item.props;
   }
   return result;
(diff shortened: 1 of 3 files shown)


aundus (approved):
Makes sense 👍

matthew-hogains (approved):

PR #808, opened 2022-09-22
docs(storybook): set actions and children control settings globally

### LINKED ISSUE #791: [storybook] promote `actions` config to top-level preview.js
we can use a regex that assigns actions based on props with a prefix `on` (see atlas storybook setup)

akdetrick:
fixes #791 

Moves some story-level configuration to the global config file:
- Actions should fire on any prop starting with `on`
- Controls for `children` prop should always be disabled in the props tables

### DIFF
--- a/src/Dialog/index.stories.js
+++ b/src/Dialog/index.stories.js
@@ -137,10 +137,6 @@ FocusManagement.parameters = {
 export default {
   title: "Components/Dialog",
   component: Dialog,
-  argTypes: {
-    children: { control: false },
-    onUserDismiss: { action: "user dismiss" },
-  },
   parameters: {
     docs: {
       page: DialogLayout,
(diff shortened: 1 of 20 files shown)


src/Dialog/index.stories.js
```diff
@@ -160,21 +156,42 @@ export const PopoverDialog = () => {
...
         `}
       </style>
-      <Popover closeOnContentClick content={
-        <div className="popover-content">
-          <div tabIndex="0" role="button" onClick={() => { setIsDialogOpen(true) }} onKeyDown={() => { }}>
-            Open Modal
-          </div>
-          <div tabIndex="0" role="button" onClick={() => { }} onKeyDown={() => { }}>
-            Does Nothing
+      <Popover
```
francesbhughes, line 159: This might be a separate issue, but when I try to close the popover open modal by clicking 'x', the modal does not close.

francesbhughes (approved):
Looks great! Left one comment but it's not a blocker

src/Dialog/index.stories.js
akdetrick: filing a follow up to investigate

Issue #807, opened 2022-09-15
[Radio Buttons]: Address design review feedback for card-kind radio buttons 


agreatscott:
Radio Buttons with `kind="card"` don't quite match the Figma - font weight should be bold and font color should be brand primary

https://www.figma.com/file/7o3se9SreiPKG4QWpGsAxg/BAO?node-id=2%3A2435

PR #817, opened 2022-10-03
fix(RadioButtons): card-style radio buttons should have font weight bold and color theme primary

agreatscott:
close https://github.com/narmi/design_system/issues/807

image


akdetrick (approved):
🙌

PR #834, opened 2022-10-25
feat(Checkbox): add error prop to display error beneath checkbox

agreatscott:
Changed the fragment to a `div` to add the `.nds-checkbox-container` classname. This is because there is a margin beneath `.nds-checkbox` label, but if an error is displayed beneath that label, we want that margin below the error. 

image

### DIFF
--- a/src/Checkbox/index.js
+++ b/src/Checkbox/index.js
@@ -14,6 +14,7 @@ const Checkbox = ({
   defaultChecked,
   checked,
   value,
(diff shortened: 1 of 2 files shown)


graphicsforge (approved):
Oh, good call... I'm shocked we hadn't noticed problems anywhere else checkboxes are used

akdetrick (approved):

PR #867, opened 2023-01-19
style(font): Line height for smallText should be 1.25

kylekennedy6:
for narmi/banking#26643

For the comment: line height on fine print is too big. should follow NDS and be 1.25

**Before:**
Screenshot 2023-01-19 at 3 07 13 PM

**After:**
Screenshot 2023-01-19 at 3 06 46 PM


kylekennedy6:
handling in banking

PR #868, opened 2023-01-19
fix(radiobuttons): Decrease padding between option and text

kylekennedy6:
for https://github.com/narmi/banking/issues/26643

For the comment: spacing btwn the button and text also seems too big (should be 28px padding)

### DIFF
--- a/src/RadioButtons/index.scss
+++ b/src/RadioButtons/index.scss
@@ -27,7 +27,7 @@
   display: inline-block;
 
   .nds-radiobuttons-option {
-    padding-left: 34px;
+    padding-left: 28px;
 
     .nds-radio {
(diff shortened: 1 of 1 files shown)


akdetrick (changes requested):
Looks good! You'll just need to amend the commit message to be `fix` instead of `style`.

Conventional commits does not trigger a release on `style` (seems wrong, but that's the spec)

kylekennedy6:
> Looks good! You'll just need to amend the commit message to be `fix` instead of `style`.
> 
> Conventional commits does not trigger a release on `style` (seems wrong, but that's the spec)

Done!

akdetrick (approved):

PR #876, opened 2023-02-01
feat(Tag): add new `grey` kind

rogerioshieh:
Result:
image


akdetrick:
Is there any meaning we can tie this to beyond its color? For example, `success`
Anything we can do to communicate design intent for this tag variant would be helpful (if possible)

rogerioshieh:
Good question. I think `neutral` would convey more meaning, what do you think?

akdetrick (approved):

PR #883, opened 2023-02-13
fix(Input): error color wins against focus color

graphicsforge:
resolves https://github.com/narmi/banking/issues/27833

was:
Screenshot from 2023-02-13 17-31-05
Is now:
Screenshot from 2023-02-13 17-22-13


akdetrick (approved):
Thanks for the fix!

Issue #882, opened 2023-02-09
[Select] Multiple selects on page always open the first


akdetrick:
### Component/Module

`Select`

### Screenshots (if applicable)
https://user-images.githubusercontent.com/231252/217864993-b3551135-b3f8-4a43-a7de-e95b5eaafb06.mov

### Reproduction Steps

- Put multiple `Select` components on a single page
- Clicking on any `Select` will open the _first_ `Select` in the DOM

### Additional Details / Context
It looks like we're passing the same `id` to all `Select` components. Removing that `id` of "nds-select" should resolve the issue.

reshav-abraham:
This wasn't happening for me locally @matthew-hogains @DerekQuoc  are you currently experiencing this?

matthew-hogains:
I can reproduce on demo2 uat. I can demo to you later this week

reshav-abraham:
This occurs when clicking the placeholder text

akdetrick:
@reshav-abraham This probably has to do with the `id` we pass into `useSelect` options. Every `Select` label is htmlFor pointing to the same `input[id='whatever']` probably

reshav-abraham:
making https://github.com/narmi/design_system/blob/fix/ddowntriggerlabel/src/Select/index.js#L73 unique did work for me

reshav-abraham:
What would be the best way to guarantee the id is unique? I noticed some implementations of using a label wrapped the component in the label

PR #888, opened 2023-02-20
fix(DropdownTrigger): use span instead of label

reshav-abraham:
closes https://github.com/narmi/design_system/issues/882

label's have a `for` and `id` attribute that we could take advantage here. I could try exploring that solution instead. I also saw that it was a common practice to wrap buttons and inputs with labels. However not all html components can be wrapped in a label tag. The easiest solution is using a span, not sure if this would break anything though

### DIFF
--- a/src/DropdownTrigger/index.js
+++ b/src/DropdownTrigger/index.js
@@ -43,9 +43,9 @@ const DropdownTrigger = React.forwardRef(
           {...otherProps}
         >
           {labelText && (
-            <label className="nds-dropdownTrigger-label" {...labelProps}>
+            <span className="nds-dropdownTrigger-label" {...labelProps}>
               {labelText}
-            </label>
+            </span>
           )}
           {displayValue && <span>{displayValue}</span>}
           {showOpenIndicator && (


src/Select/index.js
```diff
@@ -70,7 +70,7 @@ const Select = ({
   );
 
   const downshiftOpts = {
-    id: "nds-select",
+    id: `nds-select-${label}`+ Math.random().toString(16).slice(2),
```
akdetrick, line 73: This works great on the client but will unfortunately break in Next.js consumers, because the server-rendered component will not match the client rendered component.

We're only noticing this issue now because we typically only render one `Select` at a time. I'm thinking that we can allow this component to behave like it always has _unless_ an `id` prop is passed in:

`const Select = ({ id: idProp })`

```
id: idProp || `nds-select-${label}`
```

```
/** unique id attribute of the input (used for `htmlFor`) */
id: PropTypes.string.isRequired
```

akdetrick (changes requested):

src/Select/index.js
akdetrick: ☝️ This fix won't break any existing usage, but will throw warnings if `id` isn't specified, which is actually a good thing, because we want `htmlFor` to be guaranteed to work.

In the cases where we do observe the bug (multiple selects in one page), the fix will be to pass in unique `id` props

src/Select/index.stories.js
```diff
@@ -86,7 +87,7 @@ export const InAForm = () => {
           )
         </p>
       </div>
-      <Select label="Account" onChange={setInputValue}>
+      <Select id={"2"} label="Account" onChange={setInputValue}>
```
akdetrick, line 90: nitpick - I'd prefer to use non-numeric ids in the examples. For this one, something like `id="account-field"`

akdetrick (approved):
Thanks for the fix! One minor thing we should probably clean up in docs, but otherwise good to go

src/Select/index.stories.js
akdetrick: (encouraging people to name the things reduces the chances of two ids colliding again)

akdetrick (approved):

Issue #891, opened 2023-02-23
[Select] having less than 2 options crashes the page


xiexiewenting:
### Component/Module

It is possible that `Select` has less than two options (for instance, when a user has no funding cards and wishes to add one via `Select.Action`).

### Reproduction Steps

Use the `Select` component and give it only one `Select.Item` or `Select.Action`.

### Expected Result

The selector renders.

### Actual Result

Page crashes.

### Additional Details / Context

https://github.com/narmi/design_system/blob/910783221ccb53c371187b7d801c463c64368dda/src/Select/index.js#L68-L70

Issue is that when there are less than two child nodes, `children` becomes an object instead of an array and becomes very unhappy about `children.filter`

PR #892, opened 2023-02-23
fix(select): correctly render with less than two options

xiexiewenting:
resolves https://github.com/narmi/design_system/issues/891

Screen Shot 2023-02-23 at 4 44 57 PM
Screen Shot 2023-02-23 at 4 45 12 PM


akdetrick (approved):
Makes sense to filter on the array returned from `toArray` rather that trying to directly filter on `children`. Thanks!

PR #893, opened 2023-02-27
style(Dialog): Make dialogs transition open and be 120px from top

kyleschustak:
Resolves #28470

https://user-images.githubusercontent.com/51931432/221645503-5edd2b67-3121-4c19-abb3-41f67fa3d315.mov


akdetrick (changes requested):
The commit type must be one of:
> build:, chore:, ci:, docs:, style:, refactor:, perf:, test:, feat:, fix:

I recommend `refactor` for this. The commits in NDS must [conform to conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

It looks like maybe the linter didn't catch this as it's technically valid, but the release process will ignore this commit which means the version won't get bumped (`chore`, `ci`, `docs`, `test`, and custom types skip the release process)

akdetrick (approved):

PR #904, opened 2023-03-01
fix(Select): prevent chevron from overlapping other open popovers

### LINKED ISSUE #898: [Select] downward chevron of another select still visible when select is opened
### Component/Module

Select

https://user-images.githubusercontent.com/54218373/221978673-63f1cd91-1552-4ede-bb26-1271991485ae.mov

### Reproduction Steps

Put two selects on the same page (and ensure the 1st select has enough options that when expanded, cover the 2nd select).  

Expand the first select

### Expected Result

Do not see the downward chevron of the 2nd select

### Actual Result

See the downward chevron of the 1st select

Probably has to do with the fact that the trigger chevron has a z-index of 999 https://github.com/narmi/design_system/blob/7351890ad611c52eff6d53caff7ebaaafbc65958/src/DropdownTrigger/index.scss#L44

whereas the select list body has a z-index of 3
https://github.com/narmi/design_system/blob/7c39b8844a5183a2579d998556a74406e3e5381e/src/Select/index.scss#L8

akdetrick:
closes #898 

Prevent the chevron in closed Select triggers from overlapping open Select popovers

Screen Shot 2023-03-01 at 5 10 47 PM
Screen Shot 2023-03-01 at 5 10 52 PM

### DIFF
--- a/src/DropdownTrigger/index.scss
+++ b/src/DropdownTrigger/index.scss
@@ -45,7 +45,15 @@
   right: var(--space-s);
   top: 50%;
   transform: translateY(-50%);
-  z-index: 999; /* appear above popovers per design */
+
+  // only overlap popovers if in an open state
+  &.narmi-icon-chevron-up {
+    z-index: 4;
+  }
+
+  &.narmi-icon-chevron-down {
+    z-index: 1;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


xiexiewenting (approved):
very nice !! 🥳

PR #915, opened 2023-03-10
fix(Popover): only open with Enter key on trigger event

akdetrick:
fixes: https://github.com/narmi/banking/issues/29051

Instead of attaching the Enter key event to window, use only the trigger to intercept that event. Escape still works to close the popover if it's open.


xiexiewenting:
this fix makes sense !! thank you so much !!

xiexiewenting (approved):

PR #916, opened 2023-03-14
fix(Button, Checkbox, TextInput): add aria-label

graphicsforge:
related to https://github.com/narmi/banking/issues/29307

Just adding a couple aria-label attributes


graphicsforge:
@akdetrick what screen reader testing tools should we be trying out?

akdetrick (approved):

akdetrick:
@graphicsforge A combination of using Voiceover and validation tools like axe and the storybook a11y panel should be enough

PR #917, opened 2023-03-14
refactor(icons): Use icon font resources instead of inlining assets

### LINKED ISSUE #746: [icons] Find a better long-term solution for icon font serving
As a "temporary" fix for #739, we began inlining icon font assets to get them to work in a Next.js consumer production build.

Let's find a way to fix this that doesn't involve bloating our CSS to over 200kb.

akdetrick:
fixes #746 

Makes updating icons much more simple because we no longer will need to inline font assets. Also reduces bundle size.

- Update webpack rules for copying and inlining font assets
- Use the artifacts from icomoon directly (e.g. `style.css`)
- Add `fonts/` dir copied directly from icomoon (allows us to do a relative import without the `./` which doesn't work in Next.js consumers)

### Prod build in Next.js 12 consumer
Screen Shot 2023-03-14 at 7 02 06 PM

### Reviewing this PR

The README changes more or less explain what will be different moving forward.


src/icons/style.css
```diff
@@ -1,6 +1,10 @@
 @font-face {
   font-family: 'icomoon';
-  src: url('./icomoon.woff?ookznz') format('woff');
```
akdetrick, line 3: The `./` path is _magic_ in Next.js, which is why we were inlining the assets (dumping the file contents as base64 directly into the CSS).

Because this increased our CSS file size, we had an extra step for updating icons where you'd have to remove all font assets that weren't `woff`.

Now that the icon font assets are no longer inlined, we can include all of them in the `src` stack.

xiexiewenting (approved):
🥳

PR #925, opened 2023-04-12
feat(icons): inline icon font asset

akdetrick:
Fixes issue where prod builds of Nextjs were not fetching icomoon font assets correctly, resulting in CORS errors.

This webpack change causes the `woff` referenced in `src/icons/style.css` to be inlined as base64, working around the fetch entirely until we solve this problem in a more long term way.

### DIFF
--- a/src/icons/style.css
+++ b/src/icons/style.css
@@ -1,10 +1,6 @@
 @font-face {
   font-family: 'icomoon';
-  src:
-    url('fonts/icomoon.woff2?ookznz') format('woff2'),
-    url('fonts/icomoon.ttf?ookznz') format('truetype'),
-    url('fonts/icomoon.woff?ookznz') format('woff'),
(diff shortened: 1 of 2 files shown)


src/icons/style.css
```diff
@@ -1,10 +1,6 @@
 @font-face {
   font-family: 'icomoon';
-  src:
-    url('fonts/icomoon.woff2?ookznz') format('woff2'),
-    url('fonts/icomoon.ttf?ookznz') format('truetype'),
-    url('fonts/icomoon.woff?ookznz') format('woff'),
-    url('fonts/icomoon.svg?ookznz#icomoon') format('svg');
+  src: url('fonts/icomoon.woff?ookznz') format('woff');
```
akdetrick, line 3: Because we're inlining this asset as base64, I removed the other fonts from the source stack to use the widely compatible `woff` format.

This helps us keep the CSS from getting huge for now.

kyleschustak (approved):

PR #929, opened 2023-05-09
fix(dist): add deprecations json to `dist/`

akdetrick:
relates to https://github.com/narmi/banking/issues/15579

Adds a `deprecations.json` to `dist/` that lists important deprecations (things that will be removed in the next major version).

#### Why?

- a step toward a more automated deprecation strategy for NDS
- serves as a deprecation list for dangerfile rules in internal Narmi software

### DIFF
--- a/src/deprecations.json
+++ b/src/deprecations.json
@@ -1,0 +1,20 @@
+{
+  "cssCustomProperties": [
+    "--nds-black",
+    "--nds-grey",
+    "--nds-medium-grey",
+    "--nds-lightest-grey",
+    "--nds-smoke-grey",
+    "--nds-red",
+    "--nds-white",
+    "--nds-font-family",
+    "nds-header-font",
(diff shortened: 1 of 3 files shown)


src/deprecations.json
```diff
@@ -0,0 +1,20 @@
...
+    "nds-header-font",
+    "subdued-20-opacity",
+    "subdued-10-opacity"
+  ],
+  "components": [
+    "Card",
+    "Dropdown",
+    "PlainButton"
+  ]
+}
```
akdetrick, line 20: In the future, we could automate the creation of this file in `dist/` using jsDoc (e.g. the `@deprecated` tag across all js/jsx/scss files).

Worst case scenario, this file goes stale. That seems like a tolerable risk, as this is an opt-in file from `dist` and will not be included in the functional library.

matthew-hogains (approved):

PR #931, opened 2023-05-12
feat(TextInput): add support for clear input icon

rogerioshieh:
Sample:

https://github.com/narmi/design_system/assets/40327446/0ea68904-2f69-48fc-b0e8-b3dc4b6b742c

### DIFF
--- a/src/Input/index.js
+++ b/src/Input/index.js
@@ -23,6 +23,8 @@ const Input = ({
   label,
   startIconClass,
   endIconClass,
+  displayClearInputIcon,
+  clearInput,
   disabled,
   multiline = false,
   decoration,
@@ -41,6 +43,15 @@ const Input = ({
     search ? "search" : "",
   ].join(" ");
 
+  const endIconJsx = displayClearInputIcon ? (
+    <div
+      className={`nds-input-icon nds-input-close narmi-icon-x`}
+      onClick={clearInput}
+    ></div>
+  ) : (
+    <div className={`nds-input-icon ${endIconClass}`}></div>
+  );
+
   return (
     <div className={className} onClick={onClick} style={style}>
       <div className="nds-input-box">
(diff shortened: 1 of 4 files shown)


src/Input/index.js
```diff
@@ -41,6 +43,15 @@ const Input = ({
     search ? "search" : "",
   ].join(" ");
 
+  const endIconJsx = displayClearInputIcon ? (
+    <div
```
akdetrick, line 47: Let's add some props for accessibility to this:

```suggestion
    <div
       role="button"
       tabIndex="0"
       aria-label="Clear Input"
```

src/Input/index.js
```diff
@@ -72,6 +81,8 @@ Input.propTypes = {
   startIconClass: PropTypes.node,
   /** full `narmi-icon-<shape>` className for icon at end of input */
   endIconClass: PropTypes.node,
+  displayClearInputIcon: PropTypes.bool,
+  clearInput: PropTypes.func,
```
akdetrick, line 85: Should this logic be internal to `TextInput`, or is it important to give consumers direct control/responsibility over clearing behavior?

src/Input/index.js
```diff
@@ -72,6 +81,8 @@ Input.propTypes = {
   startIconClass: PropTypes.node,
   /** full `narmi-icon-<shape>` className for icon at end of input */
   endIconClass: PropTypes.node,
+  displayClearInputIcon: PropTypes.bool,
```
akdetrick, line 84: Optional/minor naming suggestion for this prop. It's an icon that does a thing, so "button" in the name may be more clear

```suggestion
  showClearButton: PropTypes.bool,
```

akdetrick (changes requested):
Looks good! Just a few suggestions for clarity and a11y

src/TextInput/index.js
```diff
@@ -106,6 +109,8 @@ TextInput.propTypes = {
   startIcon: PropTypes.oneOf(VALID_ICON_NAMES),
   /** Name of Narmi icon to place at the end of the input box */
   endIcon: PropTypes.oneOf(VALID_ICON_NAMES),
+  /** Display an X at the end of label that clears input on click. Takes precedence over endIcon when input is not empty */
```
akdetrick, line 112: > Takes precedence over endIcon when input is not empty

I was on my way here to say we should mention this. Thanks for writing good prop docs!

src/Input/index.js
akdetrick: I misread this as a prop on `TextInput` but it's actually on `Input`

akdetrick (approved):

PR #939, opened 2023-06-12
style(icons): set text-rendering: geometricPrecision;

graphicsforge:
from https://github.com/narmi/banking/issues/25927

sets the text-rendering style on icons

### DIFF
--- a/src/icons/style.css
+++ b/src/icons/style.css
@@ -6,7 +6,7 @@
   font-display: block;
 }
 
-[class^="narmi-icon-"], [class*=" narmi-icon-"] {
+[class*="narmi-icon-"] {
   /* use !important to prevent issues with browser extensions that change fonts */
   font-family: 'icomoon' !important;
   speak: never;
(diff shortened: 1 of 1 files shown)


src/icons/style.css
```diff
@@ -17,6 +17,7 @@
   line-height: 1;
 
   /* Better Font Rendering =========== */
+  text-rendering: geometricPrecision;
```
akdetrick, line 20: Could we scope this to `[class^="narmi-icon"]`? 

Chrome will render all text too light or too thin for design's liking with this set

akdetrick (changes requested):
makes sense - just needs to affect icons only

src/icons/style.css
graphicsforge: Note: we chatted out-of-band and decided to switch the selector to `[class*="narmi-icon-"]`

akdetrick (approved):

PR #940, opened 2023-06-14
feat(component): icon button

### LINKED ISSUE #930: IconButton
### Short Description
A standard `Button` that only displays an icon (with a required, accessible label).

### Mockup

[Example](https://github.com/narmi/banking/pull/31592#discussion_r1188976403)

### Responsibilities
Defaults for all props.

- Render an icon by name
- Adding an `aria-label`
- Using same size and color variant options as `Button`
- Default to `kind="plain"` (without underline)
- Padding options / click target sizing (unclear if it 

- Rendering any kind of visible label
- `as` prop (it should always render a button)
- 

### Props Signature

```
<IconButton
  size={standard sizes}
  kind={standard Button kinds}
  paddingSize?={standard size tokens}
  name={nds icon name}
  label="REQUIRED LABEL"
/>
```

### Open questions

Some of these will require a discussion with design:

- Should we support multiple variants (`kind`) of `IconButton`, or just "plain"?
- Are hover states consistent across icon buttons in design?
- Should the padding be configurable, or controlled by `size` prop?
- Do we need to have an option to expand the click target area out of document flow?

JohnathanWeisner:
resolves #930

Adds an IconButton component.

### DIFF
--- a/src/IconButton/index.js
+++ b/src/IconButton/index.js
@@ -20,7 +20,6 @@ const IconButton = ({
   kind = "plain",
   size = "m",
   testId = "nds-icon-button",
-  children,
   label,
   name,
   onClick = () => {},
@@ -66,6 +65,8 @@ IconButton.propTypes = {
   onClick: PropTypes.func,
   /** Optional value for `data-testid` attribute */
   testId: PropTypes.string,
+  /** className for adding classNames to the icon button  */
+  className: PropTypes.string,
 };
 
 export default IconButton;
--- a/src/IconButton/index.js
+++ b/src/IconButton/index.js
@@ -18,7 +18,7 @@ export const VALID_ICON_NAMES = iconSelection.icons.map(
 const IconButton = ({
   disabled = false,
   kind = "plain",
-  size = "m",
+  textSize = kind === "plain" ? "xs" : "l",
   testId = "nds-icon-button",
   label,
   name,
@@ -31,7 +31,7 @@ const IconButton = ({
         "nds-typography",
         "nds-icon-button",
         `nds-icon-button--${kind}`,
-        `fontSize--${size}`,
+        `fontSize--${textSize}`,
         {
           "nds-icon-button--disabled": disabled,
         },
@@ -53,9 +53,9 @@ IconButton.propTypes = {
   /** disables the icon button when set to `true` */
   disabled: PropTypes.bool,
   /** style of icon button to render */
-  kind: PropTypes.oneOf(["primary", "secondary", "negative", "plain"]),
-  /** size variant of icon button */
-  size: PropTypes.oneOf(["xs", "s", "m", "l"]),
+  kind: PropTypes.oneOf(["action", "plain"]),
+  /** Optional text size of the icon in the icon button defaults different for different kinds (plain/action)*/
+  textSize: PropTypes.oneOf(["xs", "s", "m", "l"]),
   /** Click callback, with event object passed as argument */
   onClick: PropTypes.func,
   /** Optional value for `data-testid` attribute */
--- a/src/IconButton/index.js
+++ b/src/IconButton/index.js
@@ -18,7 +18,7 @@ export const VALID_ICON_NAMES = iconSelection.icons.map(
 const IconButton = ({
   disabled = false,
   kind = "plain",
-  textSize = kind === "plain" ? "xs" : "l",
+  textSize = kind === "plain" ? "m" : "l",
   testId = "nds-icon-button",
   label,
   name,
--- a/src/IconButton/index.js
+++ b/src/IconButton/index.js
@@ -23,16 +23,12 @@ const IconButton = ({
   label,
   name,
   onClick = () => {},
-  className,
-  ...otherProps
 }) => {
   return (
     <button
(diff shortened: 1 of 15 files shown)


src/IconButton/index.js
```diff
@@ -0,0 +1,72 @@
...
+  testId = "nds-icon-button",
+  label,
+  name,
+  onClick = () => {},
+  className,
+  ...otherProps
+}) => {
+  return (
+    <button
+      tabIndex={0}
```
akdetrick, line 31: we get this for free with a `button` element
```suggestion
```

src/IconButton/index.js
```diff
@@ -0,0 +1,72 @@
...
+      onClick={onClick}
+      {...otherProps}
```
akdetrick, line 33: We generally avoid spread props on NDS components except for form elements. 

The idea is that we want to prevent `style` and `className` from being passed - if there's a styling change we need to make in product, we should work with design to come up with a standard variant instead of accepting arbitrary styles.

In this case, your `className` prop comes after spread props, so we may be fine if `style` is plucked from `...otherProps`

src/IconButton/index.scss
```diff
@@ -0,0 +1,100 @@
...
+  outline: inherit;
+  font-size: var(--font-size-default);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: var(--button-radius);
+  height: 32px;
+  width: 32px;
+  /* mobile buttons are larger for ease of tapping */
+  @media (max-width: #{map.get($breakpoints, "s")}) {
```
akdetrick, line 17: Optional suggestion - we have a Sass helper to do this in a mobile-first way:

```
    /* small is "default" */
    height: 48px;
    width: 48px;;
    
    /* once the viewport reaches medium OR greater, apply these styles: */
    @include atMediaUp("m") {
        height: 32px;
        width: 32px;
    }
```

src/IconButton/index.scss
```diff
@@ -0,0 +1,100 @@
...
+    height: 48px;
+    width: 48px;
+  }
+
+  /* Nudge the large size icons to be centered more accurately */
+  &.fontSize--l {
+    [class^="narmi-icon-"] {
+      position: relative;
```
akdetrick, line 25: would recommend translating instead, but not a huge deal to keep it this way

`transform: translate(-1px, -0.5px);`

Translate accomplishes the same thing without any performance hits if for some reason we want to animate hover or active states

akdetrick (commented):
This is really nice, thanks!

Before I approve, I'm going to check in with design on:
- variants needed
- design details (I agree with your choices)
- if we need to support a visual "active" state

akdetrick (changes requested):
Updates from design:

https://github.com/narmi/design_system/issues/930#issuecomment-1608110313

**tl;dr**
- `kind: PropTypes.oneOf(["action", "plain"]),`
- direct size control should be removed for the initial version of this component
- mocks/specs are in the comment linked above

src/IconButton/index.scss
```diff
@@ -0,0 +1,58 @@
...
+  &.fontSize--l {
+    [class^="narmi-icon-"] {
+      transform: translate(0px, -1px);
+    }
+  }
+}
+
+.nds-icon-button--action {
+  --button-radius: 50%;
+  color: var(--color-secondary);
```
akdetrick, line 24: Initial color should be secondary text color:

```suggestion
  color: var(--font-color-secondary);
```

src/IconButton/index.scss
```diff
@@ -0,0 +1,58 @@
...
+    color: var(--theme-primary);
+    background-color: rgba(var(--theme-rgb-primary), var(--alpha-5));
+    text-decoration: none;
+  }
+}
+
+.nds-icon-button--plain {
+  height: 24px;
+  width: 24px;
+  color: var(--theme-secondary);
```
akdetrick, line 36: Same initial color for plain buttons:

```suggestion
  color: var(--theme-secondary);
```

src/IconButton/index.js
```diff
@@ -0,0 +1,67 @@
...
+  /** Name of Narmi icon */
+  name: PropTypes.oneOf(VALID_ICON_NAMES).isRequired,
+  /** Renders the icon button label */
+  label: PropTypes.string,
+  /** disables the icon button when set to `true` */
+  disabled: PropTypes.bool,
+  /** style of icon button to render */
+  kind: PropTypes.oneOf(["action", "plain"]),
+  /** Optional text size of the icon in the icon button defaults different for different kinds (plain/action)*/
+  textSize: PropTypes.oneOf(["xs", "s", "m", "l"]),
```
akdetrick, line 58: What do you think about eliminating `xs` and maybe even `s` as valid options?

I don't think we ever want to go below "m" text, right?

akdetrick (changes requested):
Looks good! Just a few minor things before we merge

src/IconButton/index.js
JohnathanWeisner: Unfortunately the info icon is `xs` size in the figma.

akdetrick (approved):
🚀

PR #945, opened 2023-06-22
Add option for Tooltip component to be controlled with isOpen prop

JohnathanWeisner:
This allows the `Tooltip` component to be a controlled component which is needed for https://github.com/narmi/banking/issues/33260.

A developer can pass an optional `isOpen` property to control if the `Tooltip` is open or not.

### DIFF
--- a/src/Tooltip/index.js
+++ b/src/Tooltip/index.js
@@ -15,26 +15,30 @@ const Tooltip = ({
   wrapperDisplay = "inline-flex",
   maxWidth = "400px",
   testId,
+  isOpen,
 }) => {
-  const [isOpen, setIsOpen] = useState(false);
+  const isControlled = isOpen === true || isOpen === false;
+  const [open, setOpen] = useState(false);
   const delays = {
     open: 500,
     close: 100,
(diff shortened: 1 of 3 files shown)


src/Tooltip/index.js
```diff
@@ -15,26 +15,30 @@ const Tooltip = ({
   wrapperDisplay = "inline-flex",
   maxWidth = "400px",
   testId,
+  isOpen,
 }) => {
-  const [isOpen, setIsOpen] = useState(false);
+  const isControlled = isOpen === true || isOpen === false;
```
akdetrick, line 20:
```suggestion
  const isControlled = isOpen !== undefined;
```

akdetrick (approved):
Looks good - left a suggestion for the `isControlled` expression and there's one lint error to fix before merging

src/Tooltip/index.js
akdetrick: Your expression is more type safe - disregard unless you like the simple one better (checks for presence of the prop rather than value)

PR #952, opened 2023-06-26
refactor(Checkbox)!: use auto spacing between checkboxes

### LINKED ISSUE #878: [Checkbox]: checkbox should not be opinionated as to spacing around it
The NDS checkbox container has a bottom margin to it ([code here](https://github.com/narmi/design_system/blob/main/src/Checkbox/index.scss#L2)). This is unlike all the other input elements in NDS (such as `TextInput`, `DateInput`) but maybe the checkbox is intentionally different. 

It is likely written this way so that stacked checkboxes are appropriately spaced automatically, but when a checkbox is standalone, it can make spacing around it difficult. For example, if the expected space between the checkbox and the next element is 40px (`--space-xl`), it isn't possible to make this happen using only NDS spacing without overriding NDS values since it would require a stacking margin of 28px.

akdetrick:
closes #878 

**Targets `releases/v3`**

Instead of adding a margin to every checkbox container, which will create a spacing side effect with the element directly below the component, this PR makes checkbox spacing automatically appear between checkboxes based on if they directly follow each other in the DOM.

Screen Shot 2023-06-26 at 4 39 46 PM

### DIFF
--- a/src/Checkbox/index.scss
+++ b/src/Checkbox/index.scss
@@ -1,4 +1,4 @@
-// only apply top margin to a checkbox container
+// only apply top spacing to a checkbox container
 // that is preceeded by another checkbox container
 .nds-checkbox-container ~ .nds-checkbox-container {
   padding-top: var(--space-s);
--- a/src/Checkbox/index.scss
+++ b/src/Checkbox/index.scss
@@ -1,20 +1,22 @@
-.nds-checkbox-container {
-  margin-bottom: var(--space-s);
+// only apply top margin to a checkbox container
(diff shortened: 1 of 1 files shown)


src/Checkbox/index.scss
```diff
@@ -1,20 +1,22 @@
-.nds-checkbox-container {
-  margin-bottom: var(--space-s);
+// only apply top margin to a checkbox container
```
akdetrick, line 1:
```suggestion
// only apply top spacing to a checkbox container
```

agreatscott (approved):

PR #953, opened 2023-06-27
fix(TextInput): make clearInput call onChange

rogerioshieh:
Without this, TextInput fields that have an onChange won't have it triggered when the input is cleared:

https://github.com/narmi/design_system/assets/40327446/a8989951-8db7-4d5c-999c-821fe111ca13

However, it should, since it should be the same thing as manually deleting the text. Example:

https://github.com/narmi/design_system/assets/40327446/3024635e-3727-48ac-b330-086ba57e6598


akdetrick (approved):
Makes sense - thanks for the fix!

PR #954, opened 2023-06-27
feat(Popover): add alignment prop for more position control

### LINKED ISSUE #935: [Popover] Add more placement options for menu type popovers
Currently, the only placement options are in the `side` prop, and the popover will center itself within that side.

We also need to have an alignment option for menu type popovers where the popover should appear flush right or left aligned under the trigger.

akdetrick:
closes #935 

Adds `alignment` prop to `Popover`. This prop is passed through to the positioning engine to allow us to not only pick a side, but an _alignment_ for the popover.

### Example

`<Popover side="top" alignment="start" ...otherProps />`

Screen Shot 2023-06-27 at 2 10 14 PM
Screen Shot 2023-06-27 at 2 10 23 PM

### DIFF
--- a/src/Popover/index.js
+++ b/src/Popover/index.js
@@ -16,6 +16,7 @@ import useFocusTrap from "@charlietango/use-focus-trap";
  */
 const Popover = ({
   side = "bottom",
+  alignment = "center",
   content,
   children,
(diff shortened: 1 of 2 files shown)


rogerioshieh (approved):
Looks good!

PR #955, opened 2023-06-28
refactor(Select): use portalled dropdown

### LINKED ISSUE #949: [Select] needs to portal to `document.body`
### Component/Module

`Select` in a `Dialog`

### Expected Result

The `Select` dropdown overflows the `Dialog` boundaries

### Actual Result

The `Select` dropdown is rendered within the content area of the `Dialog` which is set to `overflow: scroll`

akdetrick:
closes #949 

- Replaces focus trap library with `react-focus-lock`, which is a better implementation that can follow portals
- Adds "active' state for dropdown trigger used in `Select`
- Replace static CSS positioning in `Select` with the `react-laag` positioning hook (also used in `Popover`)

### The `Select` now portals it's dropdown, allowing the menu control to overflow the `Dialog` container:
select-in-dialog

### The positioning engine (`react-laag`) now automatically flips between top/bottom to avoid getting cut off by viewport edges
select-flip

### DIFF
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { useSelect } from "downshift";
+import { useLayer } from "react-laag";
 import cc from "classcat";
 import DropdownTrigger from "../DropdownTrigger";
 import SelectItem from "./SelectItem";
@@ -66,7 +67,9 @@ const Select = ({
   testId,
 }) => {
   // The menu should only render children that have `value` or `onSelect` prop
-  const items = React.Children.toArray(children).filter(({ props }) => "value" in props || "onSelect" in props);
+  const items = React.Children.toArray(children).filter(
+    ({ props }) => "value" in props || "onSelect" in props
+  );
 
   const downshiftOpts = {
     id: id || `nds-select-${label}`,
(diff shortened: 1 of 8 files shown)


src/Select/index.js
```diff
@@ -66,7 +67,9 @@ const Select = ({
   testId,
 }) => {
   // The menu should only render children that have `value` or `onSelect` prop
-  const items = React.Children.toArray(children).filter(({ props }) => "value" in props || "onSelect" in props);
+  const items = React.Children.toArray(children).filter(
```
akdetrick, line 70: formatting only

src/Select/index.js
```diff
@@ -66,7 +67,9 @@ const Select = ({
...
+    ({ props }) => "value" in props || "onSelect" in props
+  );
```
akdetrick, line 72: formatting only

src/Select/index.js
```diff
@@ -102,50 +105,78 @@ const Select = ({
...
-            >
-              {hasSelectedItem &&
-                selectedItem.props.value === item.props.value && (
-                  <span className="narmi-icon-check fontSize--l fontWeight--bold" />
-                )}
-              {item}
-            </li>
-          ))}
-      </ul>
+      {renderLayer(
```
akdetrick, line 136: `renderLayer` creates a portal

src/Select/index.js
```diff
@@ -102,50 +105,78 @@ const Select = ({
...
     <div className="nds-select" data-testid={testId}>
       <DropdownTrigger
-        isOpen={isOpen}
+        isOpen={showMenu}
         labelText={label}
         displayValue={getSelectedItemDisplay(selectedItem)}
         labelProps={{ ...getLabelProps() }}
         errorText={errorText}
-        {...getToggleButtonProps()}
+        {...getToggleButtonProps(triggerProps)}
```
akdetrick, line 130: This is how we combine the `react-laag` element props with the `downshift` prop getters.

All the prop getters like this from `downshift` take an argument; an object of additional props to apply.

xiexiewenting (approved):
🎉

PR #960, opened 2023-07-07
fix(radio): set check icon to presentation role

kevmo314:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role

Seems like a better fit than `img`.

For https://github.com/narmi/banking/issues/32271


agreatscott (approved):

akdetrick:
Thanks @kevmo314 !

Issue #958, opened 2023-07-05
[Tabs] Allow for scrolling tabs in responsive


francesbhughes:
Create responsive tabs component that scrolls

PR #962, opened 2023-07-07
feat(tabs): Create responsive tabs component

francesbhughes:
resolves https://github.com/narmi/design_system/issues/958
responsive-tabs

### DIFF
--- a/src/Tabs/Arrow.js
+++ b/src/Tabs/Arrow.js
@@ -1,12 +1,32 @@
-import React, { useEffect, useRef } from "react";
 import PropTypes from "prop-types";
-import IconButton from "../IconButton";
+import React, { useRef } from "react";
 import { CSSTransition } from "react-transition-group";
-import cc from "classcat";
+import IconButton from "../IconButton";
 
-const Arrow = ({ direction, onClick, show }) => {
+const Arrow = ({ direction, onClick, show, isResponsive }) => {
   const arrowRef = useRef();
 
+  if (isResponsive === true) {
+    return (
+      <div className="arrow-reponsive">
+        <CSSTransition
+          nodeRef={arrowRef}
+          in={show}
+          timeout={300}
+          unmountOnExit
+          classNames="nds-tabs-arrow"
+        >
+          <div ref={arrowRef} className="nds-tabs-arrow">
+            <IconButton
+              onClick={onClick}
+              name={direction === "left" ? "arrow-left" : "arrow-right"}
+              kind="action"
+            />
+          </div>
+        </CSSTransition>
+      </div>
+    );
+  }
   return (
     <CSSTransition
       nodeRef={arrowRef}
@@ -31,6 +51,7 @@ Arrow.propTypes = {
   direction: PropTypes.string.isRequired,
   onClick: PropTypes.func,
   show: PropTypes.bool,
+  isResponsive: PropTypes.bool,
 };
 
 Arrow.displayName = "Arrow";
--- a/src/Tabs/Arrow.js
+++ b/src/Tabs/Arrow.js
@@ -1,0 +1,38 @@
+import React, { useEffect, useRef } from "react";
+import PropTypes from "prop-types";
+import IconButton from "../IconButton";
+import { CSSTransition } from "react-transition-group";
+import cc from "classcat";
+
+const Arrow = ({ direction, onClick, show }) => {
+  const arrowRef = useRef();
+
+  return (
+    <CSSTransition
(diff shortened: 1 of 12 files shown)


src/Tabs/Arrow.js
```diff
@@ -0,0 +1,59 @@
+import PropTypes from "prop-types";
+import React, { useRef } from "react";
+import { CSSTransition } from "react-transition-group";
+import IconButton from "../IconButton";
+
+const Arrow = ({ direction, onClick, show, isResponsive }) => {
+  const arrowRef = useRef();
+
+  if (isResponsive === true) {
```
JohnathanWeisner, line 9: You could do something like ```return (isResponsive && (```. The other return doesn't need to be here. We either want to have arrows reserve space or not have arrows at all.

One good rule of thumb is to use the context properties where they need to be used and not pass them down as props.

src/Tabs/TabsPanel.js
```diff
@@ -23,7 +23,11 @@ const TabsPanel = ({ children, tabId, testId }) => {
       hidden={tabId !== selectedId ? true : undefined}
       data-testid={testId}
     >
-      {children}
+      {isResponsive ? (
```
JohnathanWeisner, line 26: I like this.

JohnathanWeisner (commented):
Looks great! Just two little changes and lets get this in front of @akdetrick .

src/Tabs/TabsList.js
```diff
@@ -1,15 +1,74 @@
...
+  };
+
+  useEffect(() => {
+    if (tabsListRef.current) {
+      updateScrollButtonState();
+    }
+  }, []);
+
+  useEffect(() => {
+    const handleScrollButtonsUpdate = debounce(() => {
```
akdetrick, line 42: Good idea to debounce these calls. Another option is to use `raf-schd` (already included in NDS) to run the operation based on the next animation frame instead of on a 100ms debounce.

Example: https://github.com/narmi/design_system/blob/48d2b2f3a881a718837acc94ec5a2d24308b941f/src/Dialog/index.js#L50

src/Tabs/TabsList.js
```diff
@@ -36,16 +96,74 @@ const TabsList = ({ children, xPadding = "none" }) => {
...
-    <ul
-      role={hasPanels ? "tablist" : undefined}
-      className={`nds-tabs-tabsList list--reset padding--x--${xPadding}`}
-      onKeyDown={hasPanels ? handleKeyDown : noop}
-      tabIndex={hasPanels ? "0" : undefined}
-      data-testid="nds-tablist"
-    >
-      {children}
-    </ul>
+    <div style={{ display: "flex" }}>
```
JohnathanWeisner, line 135: Forgot to remove all of the inline styles and replace with classNames

akdetrick (changes requested):
Looks great - left some feedback

src/Tabs/TabsList.js
```diff
@@ -36,16 +95,57 @@ const TabsList = ({ children, xPadding = "none" }) => {
...
-    <ul
-      role={hasPanels ? "tablist" : undefined}
-      className={`nds-tabs-tabsList list--reset padding--x--${xPadding}`}
-      onKeyDown={hasPanels ? handleKeyDown : noop}
-      tabIndex={hasPanels ? "0" : undefined}
-      data-testid="nds-tablist"
-    >
-      {children}
-    </ul>
+    <div className="display-flex">
```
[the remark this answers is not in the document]
francesbhughes: It looks like we still need to wrap this in a flex for the arrows to align correctly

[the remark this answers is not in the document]
akdetrick: We may need to refactor the CSS to make `.display-flex` work without being a child of `nds-tabs-responsive-container` then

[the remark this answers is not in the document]
akdetrick: Looks like you already did! Resolving this convo.

src/Tabs/index.stories.js
```diff
@@ -33,6 +33,45 @@ DefaultSelectedTab.args = {
   defaultSelectedIndex: 1,
 };
 
+export const ResponsiveTabs = (args) => (
+  <div
+    style={{
+      display: "flex",
+      width: 200,
```
akdetrick, line 40: adjusting the width to be closer to something like our "s" media query

```suggestion
      width: "360px",
```

akdetrick (approved):
:

PR #969, opened 2023-07-12
feat(Typography): darken secondary text; add thin Mullish weight

### LINKED ISSUE #967: [text a11y] use a variable Mullish font; add token for 200 weight; darken secondary color
via @hillarypollak 

### Background
Our secondary text color does not have sufficient contrast below `14px`. We still want to create visual contrast even after darkening the secondary color, so the variable font will allow us to introduce a thin font weight basically for free.

### Mullish change
The best way to introduce new weights is to use the variable version of the google font for Mullish, specifying a weight range rather than explicit weights.

**This will have no impact on download size because Mullish is a variable font**

### Tokens change

Adjust secondary font color to `#6c6c6c`.
Add token for thin font weight (200).
Add helper class for thin font weight (200).

akdetrick:
closes #967 

- Darkens secondary font color to meet AA contrast guidelines for text 14px and under
- Change Mullish font to a [variable font](https://fonts.google.com/knowledge/introducing_type/introducing_variable_fonts), which gives us all weights in one download roughly the size of a single weight
- Add design token and helper class for `200` weight Mullish ("thin")

#### before
Screen Shot 2023-07-12 at 3 16 31 PM

#### after
Screen Shot 2023-07-12 at 3 16 21 PM

#### design token
Screen Shot 2023-07-12 at 3 23 38 PM

## New Mullish weight (200 aka thin)
Screen Shot 2023-07-12 at 3 23 29 PM
Screen Shot 2023-07-12 at 3 32 50 PM

### DIFF
--- a/tokens/src/font/color.json
+++ b/tokens/src/font/color.json
@@ -3,7 +3,7 @@
     "color": {
       "heading": { "value": "{color.narmi.black.value}" },
       "primary": { "value": "{color.narmi.grey.value}" },
-      "secondary": { "value": "{color.narmi.mediumGrey.value}" },
+      "secondary": { "value": "#6c6c6c" },
       "hint": { "value": "{color.narmi.lightGrey.value}" }
     }
   }
--- a/src/fonts/fonts.css
+++ b/src/fonts/fonts.css
@@ -1,4 +1,4 @@
-@import url("https://fonts.googleapis.com/css?family=Mulish:400,600,700&subset=latin");
+@import url("https://fonts.googleapis.com/css2?family=Mulish:wght@200..700&display=swap");
 
 @font-face {
   font-family: "Narmi Matiere";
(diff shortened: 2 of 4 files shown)


tokens/src/font/color.json
```diff
@@ -3,7 +3,7 @@
     "color": {
       "heading": { "value": "{color.narmi.black.value}" },
       "primary": { "value": "{color.narmi.grey.value}" },
-      "secondary": { "value": "{color.narmi.mediumGrey.value}" },
+      "secondary": { "value": "#6c6c6c" },
```
agreatscott, line 6: it's okay that this one is hardcoded but the others are variable?

akdetrick: Yep, that's fine. The aliases point to hard coded values anyway. The `mediumGrey` token is still available, but secondary text is no longer pointing at at so we can give a slightly darker value to text here

agreatscott (approved):

PR #972, opened 2023-07-18
feat(checkbox): Allow markdown

kyleschustak:
https://github.com/narmi/banking/issues/33922


rogerioshieh (changes requested):
Can we add underline to the `a` hover inside `nds-checkbox`?

```
a:hover {
      text-decoration: underline;
    }
```

Otherwise it looks good to me, though wanted to loop in @akdetrick for a second pair of eyes.

akdetrick:
We could also change the `label` prop to accept a string OR a node, so we can pass in a jsx fragment if we need to

akdetrick (approved):

PR #974, opened 2023-07-18
fix(tabslist): Tabslist scroll removal throwing error

kyleschustak:
#32294


semin-narmi (approved):

graphicsforge (approved):
race condition where the ref gets cleaned up before the useEffect cleanup.
👍 the useEffect cleanup doesn't need to fire if the page is gone

PR #981, opened 2023-07-27
feat(Select): add category grouping feature

### LINKED ISSUE #936: [Select] Add `Select.Category` that can contain `Select.Item` children
The next version of BAO has a fancier NAICS code selector:

Screen Shot 2023-06-06 at 4 32 23 PM

Ideally we can add this functionality to `Select` in this way:

```jsx
<Select>
    <Select.Category name="Agriculture" hasDisclosure>
        <Select.Item value="(naics code)">Strawberry Farming</Select.Item>
         <Select.Item value="(naics code)">Corn and Soybean Farming</Select.Item>
    </Select.Category>
</Select>
```

### Requirements

- (`hasDisclosure=false`) Categories act as titles only, without collapsible behavior
- (`hasDisclosure`) Categories can be collapsed or minified. The state of categories should reset when the dropdown closes.
- Categories should contain `Select.Item` children.
- Categories should NOT be included in the auto complete search index
- Categories are NOT values - only headers for a section and can not be selected themselves

akdetrick:
closes #936 

Adds a `Select.Category` grouping element for select items.

The actual select items are treated by downshift as a flat list of items, but the component has some helpers to determine which items are in which category so we can open and close the disclosures as the highlight/selection moves through them.

Also made some markup changes to ensure we're not invalidly nesting list items directly inside a `details` element.

### How does this work?

The opening and closing disclosures for the categories use the [native `details` html element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details). This behaves in an uncontrolled manner, but if certain conditions are met (highlighted item is in category or selected item is in category), we force it open using its `open` attribute.

### How do I review this?

Aside from looking at the code, run the branch locally. Does the select behave in the way you expect it to?

------

Screen Shot 2023-07-27 at 5 16 04 PM

Jul-27-2023 17-27-53

### DIFF
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -11,9 +11,6 @@ import SelectCategory from "./SelectCategory";
 
 const noop = () => {};
 
-// TODO: open category if current selection is in it
-// TODO: update tests
-
 /**
  * @param {Object} item a Select.Item or Select.Action component
  * @returns {Boolean} true if the item is a Select.Action
@@ -202,6 +199,17 @@ const Select = ({
     );
   };
 
+  const getDetailsProps = (categoryChildren) => {
+    let detailsExtraProps = {};
+    if (
+      isHighlightedInCategory(highlightedIndex, categoryChildren, items) ||
+      isSelectedItemInCategory(selectedItem, categoryChildren)
+    ) {
+      detailsExtraProps.open = true;
+    }
+    return detailsExtraProps;
+  };
+
   return (
     <div className="nds-select" data-testid={testId}>
       <DropdownTrigger
@@ -239,20 +247,17 @@ const Select = ({
               <details
                 key={label}
                 className="nds-select-category"
-                open={
-                  isHighlightedInCategory(
-                    highlightedIndex,
-                    categoryChildren,
-                    items
-                  ) || isSelectedItemInCategory(selectedItem, categoryChildren)
-                }
+                {...getDetailsProps(categoryChildren)} // controls open state
               >
                 <summary className="fontWeight--bold alignChild--left--center padding--x--s padding--y-xs">
-                  <span>{label}</span>
+                  <span id={`select-category-${label}`}>{label}</span>
                   <span className="nds-category-icon narmi-icon-chevron-down" />
                   <span className="nds-category-icon narmi-icon-chevron-up" />
                 </summary>
-                <ul className="list--reset">
+                <ul
+                  className="list--reset"
+                  aria-labelledby={`select-category-${label}`}
+                >
                   {categoryChildren.map((item) => renderItem(item, items))}
                 </ul>
               </details>
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
 import React from "react";
 import PropTypes from "prop-types";
 import { useSelect } from "downshift";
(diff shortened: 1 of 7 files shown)


src/Select/index.js
```diff
@@ -6,9 +7,13 @@ import cc from "classcat";
 import DropdownTrigger from "../DropdownTrigger";
 import SelectItem from "./SelectItem";
 import SelectAction from "./SelectAction";
+import SelectCategory from "./SelectCategory";
 
 const noop = () => {};
 
+// TODO: open category if current selection is in it
+// TODO: update tests
+
```
akdetrick, line 16:
```suggestion

```

akdetrick, line 16:
```suggestion
```

akdetrick, line 16:
```suggestion
```

src/Select/index.js
```diff
@@ -50,6 +52,45 @@ export const getItemByValue = (value, items) => {
   return founditem || "";
 };
 
+/**
+ * @param {Object} item an item from `items`
```
kyleschustak, line 56: Lol @ this param description

src/Select/index.js
```diff
@@ -50,6 +52,45 @@ export const getItemByValue = (value, items) => {
...
+ * @param {array} items downshift index `items`
+ * @returns {Number} index of item
+ */
+export const getItemIndex = ({ props }, items) => {
+  const values = items.map(({ props }) => props.value);
+  return values.indexOf(props.value);
+};
+
+/**
+ * @param {Object} highlightedIndex index of currently highlight item
```
kyleschustak, line 66: Looking at the code I think this is an `int` not an `Object`

src/Select/index.js
```diff
@@ -50,6 +52,45 @@ export const getItemByValue = (value, items) => {
...
+ */
+export const isHighlightedInCategory = (
+  highlightedIndex,
+  categoryChildren,
+  items
+) => {
+  if (highlightedIndex < 0) return false;
+  const highlightedValue = items[highlightedIndex].props.value;
+  const categoryValues = categoryChildren.map((child) => child.props.value);
+  return categoryValues.includes(highlightedValue);
```
kyleschustak, line 79: Might be ok, but this makes things get a bit wonky when you have 2 items with the same value
image

src/Select/index.js
```diff
@@ -50,6 +52,45 @@ export const getItemByValue = (value, items) => {
...
+/**
+ * @param {Object} selectedItem
+ * @param {Array} categoryChildren child items in a given category
+ * @returns {Boolean} if the selected item is in the given cagetory children
+ */
+export const isSelectedItemInCategory = (selectedItem, categoryChildren) => {
+  if (!selectedItem) return false;
+  const selectedValue = selectedItem.props.value;
+  const categoryValues = categoryChildren.map((child) => child.props.value);
+  return categoryValues.includes(selectedValue);
```
kyleschustak, line 91: This doesn't work in the case that the selected value is a subset of another value
image

src/Select/index.js
kyleschustak: This was because I didn't have them have different values

src/Select/index.js
akdetrick, line 56:
```suggestion
 * @param {Object} item an item from `items` passed to downshift
```

kyleschustak (approved):

PR #982, opened 2023-07-28
fix(focus traps): diable autofocus in Popover and Dialog

akdetrick:
Disables `autoFocus` feature of `FocusLock` everywhere in NDS it's used. Autofocusing the first element can lead to a frustrating experience for users, especially in regard to a11y

### DIFF
--- a/src/Dialog/index.js
+++ b/src/Dialog/index.js
@@ -38,7 +38,6 @@ const Dialog = ({
   footer,
   width = "500px",
   testId,
-  hasAutoFocus = true,
 }) => {
   const [isContentOverflowing, setIsContentOverflowing] = useState(false);
   const contentRef = useRef(null);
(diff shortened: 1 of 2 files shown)


src/Popover/index.js
```diff
@@ -27,13 +28,16 @@ const Popover = ({
...
+  const popoverContent = closeOnContentClick
+    ? React.cloneElement(content, {
+        onClick: () => {
+          if (content.onClick) {
+            content.onClick();
+          }
+          setIsOpen(false);
+        },
+      })
+    : content;
```
akdetrick, line 40: formatting only

src/Dialog/index.js
```diff
@@ -37,6 +38,7 @@ const Dialog = ({
   footer,
   width = "500px",
   testId,
+  hasAutoFocus = true,
```
akdetrick, line 41:
```suggestion
```

matthew-hogains (approved):

PR #985, opened 2023-08-02
feat(Pagination): enable fully controlled mode

### LINKED ISSUE #984: Search and pagination do not work if using APIs
### Component/Module
Pagination

https://github.com/narmi/design_system/assets/71462029/983ffdae-577b-4ea1-bc1f-3cf385ba8d31

### Reproduction Steps
1. have the total pages controlled at state initialization. 
2. Change the page to a high page number. 
3. Use some kind of API to update the stateful list object that controls the totalPage attribute. 
4. Search where some result has a total number of pages less than the previous total pages. The state of the Pagination component will try to automatically set the previously selected page index of some high number and it. 

### Expected Result
if index is higher than expected, the pagination should reset to the defaultIndex?

### Actual Result
error
The error 
```Uncaught TypeError: Cannot read properties of undefined (reading 'indexOf')```
should be related to the previous index on reload being out of the range of the current total pages. It is interesting that the error is undefined reading indexOf. This seems like the culprit
 
https://github.com/narmi/design_system/blob/6a89038eb46bd2119c0827f8f5fb7ff0d2259765/src/Pagination/index.js#L36-L39

setting `visiblePages` to the first that includes the current state of `selectedPage` cause an undefined to return since the selectedPage would be too high. Then getting `indexOf` would fail.

### PR DESCRIPTION
**Pull request by akdetrick**
closes #984 

Refactor `Pagination` to allow fully controlled selected page and total pages.

When the selected page becomes out of bounds (most likely because total pages change), we move selection back to the beginning. If the currently selected page is within bounds, we keep that page selected.

Aug-02-2023 13-08-24

### DIFF
--- a/src/Pagination/index.js
+++ b/src/Pagination/index.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import PropTypes from "prop-types";
 import cc from "classcat";
 import Row from "../Row";
@@ -16,7 +16,7 @@ const MAX_VISIBLE_PAGES = 5;
  */
 export const _getAttributes = (
   totalPages,
-  selectedPage = 1,
+  selectedPage,
   segmentSize = MAX_VISIBLE_PAGES
 ) => {
   // create a list of page segments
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Pagination/index.js [hunk 1]
```diff
@@ -70,43 +71,57 @@ const Pagination = ({
...
+
   const {
     visiblePages,
     selectedIndex,
     showPrev,
     showNext,
     firstPage,
     lastPage,
-  } = _getAttributes(totalPages, selectedPage);
+  } = paginationAttributes;
```
akdetrick, line 102: This gets updated on first render, and any time `totalPages`, `selectedPageControlled`, or `selectedPage` (uncontrolled) changes

src/Pagination/index.js
```diff
@@ -48,7 +49,7 @@ export const _getAttributes = (
 
   return {
     visiblePages,
-    selectedIndex,
+    selectedIndex: selectedIndex >= 0 ? selectedIndex : 0,
```
akdetrick, line 52: `selectedIndex` here refers to the index of visible page numbers (the five we show in the middle).

When the selected page is out of bounds of the visible pages, we can assume it's also out of bounds of total pages and default the selection to index `0` (page 1).

matthew-hogains (approved):

PR #990, opened 2023-08-04
feat(Select): add getTypeaheadString prop for custom typeahead behavior

### LINKED ISSUE #989: [Select] add feature to customize typeahead behavior
[This PR](https://github.com/narmi/design_system/pull/981) added the `searchValue` prop for additional control. This works fine if you only have one string you'd like to search against when performing the typeahead searching, but if we want to support a more complex behavior, we'll need to give users control over downshift's `itemToString` funciton.

### NAICS codes example
We want to allow users to type _either_ a sector name or an id code to typeahead on the NAICS selector.
The new feature would work something like this:

```jsx
<Select
  label="NAICS Code"
  getTypeaheadString={(userInput, selectItem) => {
      let searchString;
      if (/\d/.test(userInput)) {
          searchString = selectItem.props.value; // user is searching by code
      } else {
          searchString = selectItem.props.searchValue || selectItem.props.value; // user is searching by name
      }
      return searchString;
  }}
>
  <Select.Item value="111110" searchValue="Soybean farming">
      Soybean farming - 111110
  </Select.Item>
```

#### caveats / unknowns

The `itemToString` method only has an argument for the item. How do we get the `userInput` argument from downshift? It's probably possible, but unclear how as of time of filing this ticket

akdetrick:
closes #989 

Enables custom typeahead behavior via the `getTypeaheadString` function prop.

### DIFF
--- a/src/Select/index.stories.js
+++ b/src/Select/index.stories.js
@@ -136,6 +136,43 @@ CustomTypeahead.parameters = {
   },
 };
 
+export const ChangingTypeaheadBehavior = Template.bind({});
+ChangingTypeaheadBehavior.args = {
+  label: "Select an Industry",
+  getTypeaheadString: (userInput, selectItem) => {
+    console.info({ userInput });
+    console.info({ selectItem });
+
+    let searchString =
+      selectItem.props.searchValue || selectItem.props.value || "";
+
+    // user is searching by code
+    if (/\d/.test(userInput)) {
+      searchString = selectItem.props.value;
+    }
+
+    return searchString;
+  },
+  children: [
+    { name: "Agriculture", code: "12345" },
+    { name: "Manufacturing", code: "55555" },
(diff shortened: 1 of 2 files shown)


src/Select/index.stories.js
```diff
@@ -136,6 +136,43 @@ CustomTypeahead.parameters = {
...
+    let searchString =
+      selectItem.props.searchValue || selectItem.props.value || "";
+
+    // user is searching by code
+    if (/\d/.test(userInput)) {
+      searchString = selectItem.props.value;
+    }
+
+    return searchString;
+  },
```
akdetrick, line 155: It's not a true fuzzy search, but exposing this function to consumers to figure out what to do is the easiest, lowest risk change to allow typeahead by both NAICS code and sector name

kyleschustak (approved):
Small issue with the storybook IDs being the same. Otherwise LGTM

PR #994, opened 2023-08-08
fix(Button): use implicit type; add 'reset' type

akdetrick:
Fix for v3 - instead of defaulting to `button` type, only set a value for `type` if a prop is explicitly passed by a consumer.

**From MDN:**
> [`submit`] is the default if the attribute is not specified for buttons associated with a `<form>`, or if the attribute is an empty or invalid value.


xiexiewenting (approved):
reasonable approach!

PR #1017, opened 2023-08-29
feat(DateInput): add `disablePortal` for use inside popover

### LINKED ISSUE #1010: [Popover] DateInput in popover causes it to close after clicking on calendar selector
### Component/Module
Popover

https://github.com/narmi/design_system/assets/71462029/2975aa63-9bf3-4660-8af4-a8bef5f97ffa

### Reproduction Steps
1. create popover with content include DateInput with calendar selector for date
2. select date from calendar selector
### Expected Result
popover stays open
### Actual Result
popover closes
probably want some way to override if the popover is open. The issue would be how to still trigger if the user clicks off of the popover.

akdetrick:
fixes #1010 

Adds `disablePortal` prop to `DateInput` to append the calendar _inside_ its parent instead of `document.body`. This is useful when we want to use a `DateInput` inside a `Popover`.

Screen Shot 2023-08-29 at 6 51 50 PM
Screen Shot 2023-08-29 at 6 51 59 PM

### DIFF
--- a/src/DateInput/index.js
+++ b/src/DateInput/index.js
@@ -58,7 +58,6 @@ const DateInput = ({
 
   useEffect(() => {
     if (disablePortal) {
-      console.dir({ ...props });
       flatpickrOptions.static = true;
     }
     flatpickr(input.current, flatpickrOptions);
--- a/src/DateInput/index.js
+++ b/src/DateInput/index.js
@@ -21,6 +21,7 @@ const DateInput = ({
   useIsoOnChange = true,
   disableMobile = false,
   testId,
+  disablePortal,
   ...props
 }) => {
   const input = useRef();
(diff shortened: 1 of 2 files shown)


src/DateInput/index.js
```diff
@@ -56,21 +57,26 @@ const DateInput = ({
   };
 
   useEffect(() => {
+    if (disablePortal) {
+      console.dir({ ...props });
```
akdetrick, line 61:
```suggestion
```

src/DateInput/index.js
```diff
@@ -56,21 +57,25 @@ const DateInput = ({
   };
 
   useEffect(() => {
+    if (disablePortal) {
+      flatpickrOptions.static = true;
```
matthew-hogains, line 61: Console statement

src/DateInput/index.js
```diff
@@ -56,21 +57,25 @@ const DateInput = ({
...
-        ref={input}
-        type="date"
-        required
-        placeholder={props.label}
-        data-testid={testId}
-        {...props}
-      />
-    </Input>
+    <div>
+      <Input {...props} label={props.label}>
```
akdetrick, line 68: static mode of flatpickr does some weird things with the input placeholder - moving the label prop to the `Input` component makes things behave more normal regardless of the flatpickr mode

src/DateInput/index.js
akdetrick: fixed

PR #1019, opened 2023-08-30
feat(ResponsiveFlex): add useViewportSizes hook and ResponsiveFlex

### LINKED ISSUE #1006: [FlexLayout] responsive helper for changing flex properties based on MQ matching
### Short Description

A layout component that makes it easier to author layouts that change flex direction, order, and alignment based on viewport size, without the need to write any new CSS.

Something with props like:
```jsx
  /** sets flex-direction (along with either 100% height or width) */
  direction?: 'row' | 'column';

  /** switches flex-direction at a given breakpoint */
  switchDirection?: 'xs' | 's' | 'm' | 'l' | 'xl';

  /** sets standard justify-content */
  justify?: 'end' | 'center' | 'spaceBetween' | 'spaceAround';

  /** sets standard align-items */
  align?: 'start' | 'end' | 'center' | 'stretch';

  /** When set, gutters are removed from `FlexItem` children */
  noGutters?: boolean;

  /** When set, `FlexItem` order is rendered in reverse */
  reverse?: boolean;

  /** React children (should be of type `FlexItem`) */
  children: React.ReactNode;
```

### Responsibilities

- Act as a flex parent
- Apply different flex properties based on props passed in
- CSS that responds to viewport size

- Spacing of the flex items themselves
- Anything other than layout

### Props Signature

TBD

akdetrick:
closes #1006 

- Adds a `ResponsiveFlex` layout helper component.
- Adds `useViewportSizes` hook. Docs could use some work, but would rather do that in a follow up

## useViewportSizes

Uses `matchMedia` API so the browser doesn't have to do any reflow work; it uses the same codepath in the browser as CSS media query evaluation.

`const {s,m,l,xl} = useViewportSizes()`

Screen Shot 2023-08-30 at 1 06 54 PM

## ResponsiveFlex

Uses the `useViewportSizes` hook to calculate flex properties based on props passed in and current viewport matches.

Screen Shot 2023-08-30 at 1 01 19 PM
Screen Shot 2023-08-30 at 1 01 11 PM

Aug-30-2023 13-04-53


src/ResponsiveFlex/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+  if (viewportMatches[toColumnAt]) {
+    flexDirection = "column";
+  }
+
+  // must be the final override, as we treat reverse as a separate concept from row/column
+  if (viewportMatches[reverseAt]) {
+    flexDirection = `${flexDirection}-reverse`;
+  }
+
+  console.info(flexDirection);
```
akdetrick, line 31:
```suggestion
```

src/hooks/useViewportSizes/index.jsx
```diff
@@ -0,0 +1,29 @@
+import { useMediaQuery } from "usehooks-ts";
```
akdetrick, line 1: Well tested, small hook that includes a lot of nice things, like listening to the MediaQueryList `onchange` events instead of throttling browser resize, among other things

src/hooks/useViewportSizes/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+  }
+) => {
+  const s = useMediaQuery(queries.s);
+  const m = useMediaQuery(queries.m);
+  const l = useMediaQuery(queries.l);
+  const xl = useMediaQuery(queries.xl);
+  return { s, m, l, xl };
+};
+
+export default useViewportSizes;
```
akdetrick, line 29: I'm starting to rethink this name... Maybe `useBreakpoints` is better

JohnathanWeisner (approved):
Wonderful addition to the design system. Nice work.

PR #1020, opened 2023-08-31
fix(icon): remove color from filter-2

henderson2350:
part of #35947

I need to make the icon white, so need to remove the color specification here.


akdetrick (approved):
Thank you!

PR #1023, opened 2023-09-07
feat(Combobox): add searchValue to items for typeahead control

akdetrick:
relates to:
- #1018 

Adds `searchValue` to combobox items, allowing developers to explicitly set the string that typeahead behavior matches against.

Screen Shot 2023-09-07 at 5 42 54 PM

PR #1024, opened 2023-09-09
feat(Combobox): add categories feature 

### LINKED ISSUE #1018: [Combobox] add all Select features
Recently `Select` had a few important features added. We should also bring those into `Combobox`.

### Definition of Done

`Combobox` has...

- [ ] Custom search string value
- [ ] Folding categories
- [ ] Typeahead behavior customization

akdetrick:
closes #1018 

**blocked by: https://github.com/narmi/design_system/pull/1023**

Adds `Combobox.Category` support, similar to the category feature in `Select`.

## How it works

When categories are passed into `Combobox`, we extract all the children as a flat list, [similar to `Select`](https://github.com/narmi/design_system/blob/a6f155ba42c9d1edb087b326b35b582487be78c1/src/Select/index.js#L129). This allows the downshift `useCombobox` hook to ignore categories and behave the same as a basic Combobox.

Categories are used to nest items inside `details` elements when rendering, [similar to how it works in `Select`](https://github.com/narmi/design_system/blob/a6f155ba42c9d1edb087b326b35b582487be78c1/src/Select/index.js#L282).

### Category rendering rules

A category is only rendered when:
- It has at least one item that has not been filtered out by the typeahead behavior

A category is only opened when:
- User clicks on the category name
- Highlight via mouse or keyboard nav moves onto an item included in the category
- When items are being filtered, so users can see the category names _and_ the suggestions while filtering

## Demo
Sep-08-2023 17-35-19

### DIFF
--- a/src/Combobox/ComboboxItem.js
+++ b/src/Combobox/ComboboxItem.js
@@ -15,7 +15,7 @@ ComboboxItem.propTypes = {
    */
   value: PropTypes.string.isRequired,
   /**
-   * Custom typeahead string. By default typeahead uses `value`.
+   * Custom typeahead string. By default, typeahead uses `value`.
    * Use this prop to pass in a custom string you'd like the user to search
    * against when using typeahead.
    */
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -80,6 +80,17 @@ export const getVisibleChildrenByCategory = (
 };
 
 /**
+ * @param {Array} items all selectable Combobox.Item children
+ * @param {String} inputValue lowercase value of input
+ * @returns {Array} Combobox.Item children, filtered by the input value
+ */
+export const defaultFilterItemsByInput = (items, inputValue) =>
+  items.filter((item) => {
+    const query = item.props.searchValue || item.props.value;
+    return query.toLowerCase().startsWith(inputValue);
+  });
+
+/**
  * Autocomplete input component following the accessible
  * [ARIA combobox pattern](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role).
  *
@@ -94,6 +105,7 @@ const Combobox = ({
   onChange = noop,
   onInputChange = noop,
   inputValue: controlledInputValue,
+  filterItemsByInput = defaultFilterItemsByInput,
   children,
   disableFiltering = false,
   errorText,
@@ -145,10 +157,10 @@ const Combobox = ({
       // Typeahead behavior - we adjust the list of available options passed
       // into `useCombobox` by filtering the initial items list from input value
       if (!disableFiltering) {
-        const filteredItems = items.filter(isSelectable).filter((item) => {
-          const query = item.props.searchValue || item.props.value;
-          return query.toLowerCase().startsWith(inputValue.toLowerCase());
-        });
+        const filteredItems = filterItemsByInput(
+          items.filter(isSelectable),
+          inputValue.toLowerCase()
+        );
         setDisplayedItems(filteredItems);
       }
 
@@ -377,6 +389,12 @@ Combobox.propTypes = {
    * as the user types.
    */
   disableFiltering: PropTypes.bool,
+  /**
+   * Optionally pass a function to customize filtering behavior
+   *
+   * Signature: `(items, inputValue) => [...filteredItems]`
+   */
+  filterItemsByInput: PropTypes.func,
   /**
    * Error message.
    * When passed, this will cause the input to render in error state.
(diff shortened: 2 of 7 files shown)


src/Combobox/index.js
```diff
@@ -107,29 +173,115 @@ const Combobox = ({
...
+    useLayer({
+      isOpen,
+      overflowContainer: true,
+      auto: true,
+      snap: true,
+      placement: "bottom-start",
+      possiblePlacements: ["top-start", "bottom-start"],
+      triggerOffset: -3,
+      containerOffset: 16,
+    });
```
akdetrick, line 187: formatting only

src/Combobox/index.js
```diff
@@ -179,58 +333,21 @@ const Combobox = ({
             "nds-combobox-list",
             "list--reset bgColor--white border--right bottom border--left",
             {
-              "nds-combobox-list--active": isOpen,
+              "nds-combobox-list--active": isOpen && displayedItems.length > 0,
```
akdetrick, line 336: Fixes existing bug where the popup still rendered when empty

src/Combobox/ComboboxItem.js
```diff
@@ -14,6 +14,12 @@ ComboboxItem.propTypes = {
    * value starting with `n`.
    */
   value: PropTypes.string.isRequired,
+  /**
+   * Custom typeahead string. By default typeahead uses `value`.
```
akdetrick, line 18:
```suggestion
   * Custom typeahead string. By default, typeahead uses `value`.
```

src/Combobox/index.js
```diff
@@ -26,6 +28,57 @@ export const isSelectable = (component) => {
...
+  let result = false;
+
+  // an item in the category is currently highlighted
+  if (highlightedIndex > -1 && displayedItems.length > 0) {
+    const highlightedValue = displayedItems[highlightedIndex].props.value;
+    const categoryValues = categoryChildren.map((child) => child.props.value);
+    result = categoryValues.includes(highlightedValue);
+  }
+
+  // user is actively filtering; open all categories and show suggestions
```
akdetrick, line 53:
```suggestion
  // user is actively filtering; default all categories to open
```

kyleschustak:
Pulled down this branch and did some testing. Here are some thoughts.

- Categories aren't highlighting on hover.
- When I expand a category it's changing from being below the box to above:
image
image
- When I select something, click off, then click back on, it treats it like a text input and nothing opens up
- I can just type some random text in the box and it seems like it's valid:
image
- It's searching just by the beginning of the string, so we can't search by name of the category _and_ number for the category, it's one or the other

akdetrick:
Added some changes to address some of these issues.

> - Categories aren't highlighting on hover.

Categories now change background on CSS mouse hover. Categories will still be skipped when navigating via keyboard.

---

> - When I expand a category it's changing from being below the box to above:

Set a preferred side for top/bottom placement in the `useLayer` hook. This is otherwise expected behavior if there's not enough room on the page, but the switch to top positioning should happen less often.

❓ Is this considered a blocker? If so, I could add `onOpen` and `onClose` callbacks you could use to add some extra space on a parent element somewhere to guarantee there's room below the input.

---

> - When I select something, click off, then click back on, it treats it like a text input and nothing opens up

**Fix in progress** Will update this comment when I push it up

---

> - I can just type some random text in the box and it seems like it's valid:

Validation is a job for the parent component, and should probably be handled by showing an error that the value isn't valid. We don't want to reject input from the component as it would be a poor UX (too short of a timeout to reject means the input clears while the user is typing/correcting; too long a timeout means that the user thinks they've completed the field for the value to disappear if it's invalid)

---

> - It's searching just by the beginning of the string, so we can't search by name of the category _and_ number for the category, it's one or the other

Fixed via the new `filterItemsByInput` prop that takes a function
Screenshot 2023-09-20 at 4 00 51 PM

src/Combobox/index.stories.js
```diff
@@ -150,6 +150,82 @@ export const InADialog = () => {
...
+      </Combobox.Item>
+      <Combobox.Item searchValue="Main Checking" value="checking2">
+        Main Checking - 13989
+      </Combobox.Item>
+      <Combobox.Item searchValue="Joint Checking" value="checking3">
+        Joint Checking - 14857
+      </Combobox.Item>
+    </Combobox.Category>,
+    <Combobox.Category label="Savings">
+      <Combobox.Item searchValue="Business Checking" value="savings1">
```
kyleschustak, line 170: Minor, but think there's some copy/pasta here. Think you meant "Business Savings"

kyleschustak (approved):

PR #1025, opened 2023-09-11
feat(SeparatorList): add noWrap prop

### LINKED ISSUE #611: [SeparatorList] inline block list breaks on spaces/separators
### Component/Module
`SeparatorList`

### Reproduction Steps
Make a long line of items in a `SeparatorList`

### Expected Result
The list does not wrap on spaces/separators

### Actual Result
The list will wrap, with the trailing separator dangling on the first line 

We should add a `noWrap` prop option defaulted to `false`. Users can enable this option for special cases like a SeparatorList inside a `Row.Item shrink`

akdetrick:
fixes #611 

Adds `noWrap` prop, which defaults to `false` to avoid breaking existing usage. This is useful when we have a SeparatorList in a Row.Item that is set to shrink - we don't always want to make the list wrap.

Screenshot 2023-09-11 at 4 27 55 PM


xiexiewenting (approved):
:tada:

PR #1039, opened 2023-09-27
fix(DateInput): use TextInput component instead of native input directly

### LINKED ISSUE #1038: [DateInput] Label only appears after a value is supplied
### Component/Module
`DateInput`

Screenshot 2023-09-26 at 2 23 51 PM

Screenshot 2023-09-26 at 2 23 57 PM

### Reproduction Steps

1. Add `altFormat="m/d/Y"` and `altInput` as props to `DateInput`
2. Add a `label` prop

### Expected Result

Expect to see the label when there is no value

### Actual Result

Do not see a value unless a date is selected

this behavior might have started after https://github.com/narmi/design_system/pull/1017

akdetrick:
fixes #1038 

Convert `input` to `TextInput` to ensure stable label rendering behavior

### DIFF
--- a/src/DateInput/index.js
+++ b/src/DateInput/index.js
@@ -1,6 +1,6 @@
 import React, { useEffect, useRef } from "react";
 import PropTypes from "prop-types";
-import Input from "../Input";
+import TextInput from "../TextInput";
 import { english } from "flatpickr/dist/l10n/default";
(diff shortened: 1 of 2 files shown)


graphicsforge (approved):
Right, Input says that it's deprecated and that TextInput should always be used

👍

PR #1040, opened 2023-09-27
fix: use consistent text-rendering for input labels

akdetrick:
relates to https://github.com/narmi/banking/issues/36941

**Ensures text rendering is consistent between faux inputs and real inputs**

We set all our body text to `text-rendering: geometricPrecision`, which does a bit more subpixel antialiasing, making the font weight appear slightly thinner in most browsers.

Input placeholders and labels however, have user agent styling of `text-rendering: auto` in most browsers.

Because `DropdownTrigger` isn't a real input, we were seeing some weirdness in forms where label thickness didn't look consistent.


akdetrick:
Minimal risk; fixes a long standing design issue

Issue #736, opened 2022-05-26
[formatNumber] Support signDisplay option


ejrb:
Money increments and decrements, such as in transaction lists, should be signed for both +ve and -ve values:

![Screen Shot 2022-05-26 at 10 41 33 AM](https://user-images.githubusercontent.com/5270155/170511555-21cac0e4-bbbc-46d9-be77-6943d5f5ebca.png)

`formatNumber` should support this option

### Technical details
- Intl.NumberDisplay supports this via the signDisplay  option with either `"always"` or `"exceptZero"` ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat))

reshav-abraham:
closed via https://github.com/narmi/design_system/pull/1042

PR #1042, opened 2023-09-28
feat(formatNumber): Add `signDisplay` as prop to `formatNumber` to use native `Intl.NumberFormat` `signDisplay` functionality 

reshav-abraham:
part of https://github.com/narmi/design_system/issues/736

Allows `signDisplay` to be passable to `Intl.NumberFormat`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat

signDisplay
When to display the sign for the number. Possible values are:

"auto" (default)
Sign display for negative numbers only, including negative zero.

"always"
Always display sign.

"exceptZero"
Sign display for positive and negative numbers, but not zero.

"negative"
Sign display for negative numbers only, excluding negative zero.

"never"
Never display sign.


akdetrick (approved):

src/formatters/formatNumber.js
```diff
@@ -32,14 +32,23 @@ const formatNumber = (input, style = "currency") => {
...
   }
 
+  const validSignDisplays = ["auto", "never", "always", "exceptZero"];
+  if (!validSignDisplays.includes(signDisplay)) {
+    throw new Error(
+      `formatNumber: invalid signDisplay argument. Must be one of: ${JSON.stringify(
+        validSignDisplays
+      )}`
+    );
+  }
```
akdetrick, line 42: thanks for the value validation here

Issue #1014, opened 2023-08-26
New: Sidebar component


matthew-hogains:
We may want to consider a different interface for this component to make it easier to reuse. This component has a lot of responsibilities that we could move to the parent in order to make the sidebar more purely presentational:

- managing history
- managing active step
- rendering the content area

Usage of a more purely presentational sidebar might look more like this:

```jsx
<Row>
   <Row.Item shrink>
      <Sidebar>
         <Sidebar.Step onClick={} key={key} active={true}>Display value</Sidebar.Step>
         <Sidebar.Step onClick={} key={key}>Display value</Sidebar.Step>
         <Sidebar.Step onClick={} key={key}>Display value</Sidebar.Step>
      </Sidebar>
   </Row.Item>
   <Row.Item>
      {/* main content */}
   </Row.Item>
</Row>
```

This would account for minor design variations (e.g. if we want to add icons to display values on only one page) without the Sidebar component collecting more and more props and complexity over time.

The parent is also managing history here, so the exact behavior on click can be customized to our needs.

_Originally posted by @akdetrick in https://github.com/narmi/banking/pull/35670#discussion_r1304576535_

akdetrick:
### Responsibilities


**This component should be responsible for the following:**
- Navigation landmark
- render nav items of any length
- show "current" selection
- hover states
- `kind` prop to handle `nav` vs `steps` variants of Sidebar

**This component should NOT be responsible for...**
- the background color it lives inside
- outer borders
- history management
- routing logic

### Props Signature


```
<Sidebar>
         <Sidebar.Step onClick={} key={key} active={true}>Display value</Sidebar.Step>
         <Sidebar.Step onClick={} key={key}>Display value</Sidebar.Step>
         <Sidebar.Step onClick={} key={key}>Display value</Sidebar.Step>
</Sidebar>
```

PR #1048, opened 2023-10-03
feat(Sidebar): add Sidebar component

cs-paz:
closes https://github.com/narmi/design_system/issues/1014

Added the sidebar component

Removed hardcoded `color`'s from `./src/icons/style.css`


src/index.scss
```diff
@@ -67,6 +67,7 @@
 @import "Pagination/";
 @import "Select/";
 @import "SeparatorList/";
+@import "Sidebar/";
```
akdetrick, line 70: I forgot, we also need to add an export to `src/index.js` to make the component available to consumers of `@narmi/design_system`

akdetrick (changes requested):
Found just one minor thing we missed, but otherwise LGTM

src/index.scss
cs-paz: Added this!

akdetrick (approved):

PR #1049, opened 2023-10-04
feat(Tabs): make responsive Tabs scrollable with no arrows

francesbhughes:
Screenshot 2023-10-04 at 11 53 01 AM
Screenshot 2023-10-04 at 11 54 45 AM
Screenshot 2023-10-04 at 2 29 17 PM

responsive_scrolling_tabs_web

responsive_scrolling_tabs_mobile


src/Tabs/TabsList.js
```diff
@@ -20,44 +16,52 @@ const TabsList = ({ children, xPadding = "none" }) => {
...
-    const nextShowLeftArrow = scrollLeft > 1;
-    const nextShowRightArrow = scrollLeft < scrollWidth - clientWidth - 1;
-
-    setShowLeftArrow(nextShowLeftArrow);
-    setShowRightArrow(nextShowRightArrow);
-    setIsResponsive(nextShowLeftArrow || nextShowRightArrow);
+    setIsResponsive(
+      scrollLeft > 1 || scrollLeft < scrollWidth - clientWidth - 1
+    );
+    const tabsSize = getTabsSize();
```
JohnathanWeisner, line 28:
```suggestion
    const allFullyVisibleTabsWidth = getAllFullyVisibleTabsWidth();
```

While I like short names too. A more accurate name would allow for a lot less digging to understand the code.

JohnathanWeisner:
What you have looks great.
I have noticed a functionality issue with the design.
You cannot click and drag to scroll on desktop when then menu is in responsive mode.


The design's inspiration has a more dropdown option to cover this case except for the smallest of screen sizes.


@hillarypollak Would love some direction on this. We could either stick with arrows or build this collapsible menu functionality.

hillarypollak:
i think the collapsible menu would be great. do you need a design for it?

JohnathanWeisner:
@hillarypollak A design would be great. To ensure we're on the same page we'll have 3 states for the responsive tabs menu.

1. Full width and all tabs visible
2. Desktop screen size and some tabs are not visible turn to drop down menu for tabs that cannot fit.
3. Mobile screen size touch scroll and no menu.

akdetrick:
@hillarypollak I think both of these options would be a good solution to avoid a frustrating UX. To add a bit of context for each option...

### Collapsible menu (like booking.com)
Something we should consider when designing for this approach is which tabs are most important to be visible to the user and available at one touch (as opposed to hidden behind a dropdown).

We may need to build in the ability to declare which tabs are ok to relocate into a dropdown at smaller sizes, and will need to specify that in all current usage of tabs and in future designs. Either that, or we can make the assumption that the first few tabs are the most important in all cases.

### Arrows
While not the most visually pleasing, it would be the easiest to implement and tabs could remain in the same order as they were designed.

hillarypollak:
I'd be okay with just keeping the first few items and collapsing the rest vs creating logic around which are the most important. I'll work on a design and loop you in for feedback once it's ready

hillarypollak:
https://www.figma.com/proto/ixoS3enxkIhoz9c9akC2O5/settings?page-id=504%3A2523&type=design&node-id=3521-1277&viewport=-716%2C-2797%2C0.32&t=XTEFs2zqOgKDfu3Q-1&scaling=min-zoom&starting-point-node-id=3521%3A1277&show-proto-sidebar=1&mode=design

let me know what you think of this

akdetrick:
@hillarypollak The more dropdown makes sense to me. I just got an idea for how we can control which tabs collapse into the more dropdown, so the tab order tradeoff probably isn't that big of a concern. 

cc @JohnathanWeisner @francesbhughes 

One way we can simplify the implementation is make the consumer responsible for what gets rendered in the dropdown.

Instead of writing logic to measure the width of each tab against the available width (a lot of reflows unless memoized against label values), the consuming application can use the [`useBreakpoints hook`](https://narmi.github.io/design_system/?path=/docs/hooks-usebreakpoints--overview) to determine when to put a specific tab into the more dropdown.

One way to do this is to create a `Tabs.MoreDropdown` equivalent to `Tabs.List`:

```jsx
<Tabs>
   <Tabs.List>
      <Tabs.Tab />
      <Tabs.Tab />
   </Tabs.List>
   <Tabs.Dropdown>
      <Tabs.Tab />
      <Tabs.Tab />
   </Tabs.Dropdown>
   {/* panels or no panels */}
</Tabs>
```

In combination with the `useBreakpoints` hook from NDS, they could just decide when to move tabs around based on design requirements and/or label length.

Here's an example where we keep the most important tabs visible in the tabs list.
```jsx
const ResponsiveTabsForFeature = () => {
   const { m } = useBreakpoints();

   return (
      <Tabs>
         <Tabs.List>
            <Tabs.Tab label="Most Important" />
            {m && (
              <Tabs.Tab label="Less important" />
            )}
            <Tabs.Tab label="Important" />
            {m && (
              <Tabs.Tab label="Not Important" />
            )}
         </Tabs.List>
         <Tabs.MoreDropdown>
            {!m && (
              <Tabs.Tab label="Less important" />
              <Tabs.Tab label="Not Important" />
            )}
         </Tabs.MoreDropdown>
      </Tabs>
   );
}
```

It's extra annoying boilerplate, but it _is_ flexible. We can decide which tabs are important to keep visible. Extra breakpoints can be used to get more granular with rules on which tabs move to the dropdown at which sizes.

**There might be a better way to solve this; just throwing out the suggestion**

JohnathanWeisner:
> Instead of writing logic to measure the width of each tab against the available width (a lot of reflows unless memoized against label values), the consuming application can use the [`useBreakpoints hook`](https://narmi.github.io/design_system/?path=/docs/hooks-usebreakpoints--overview) to determine when to put a specific tab into the more dropdown.


Certainly, the `useBreakpoints` hook presents a viable solution; however, it may not sufficiently address all scenarios. We must consider that some tabs are conditional, displayed only when specific feature flags are enabled, which would necessitate creating numerous breakpoint sets to accommodate the various permutations. A more robust approach might be to implement an automated, memoized system that dynamically allocates tabs to the dropdown menu upon resizing, thereby ensuring seamless adaptation to all possible configurations.

PR #1051, opened 2023-10-04
Style/1044 combobox design feedback

### LINKED ISSUE #1044: [Combobox] Design feedback
Via Slack (Hillary)

> - Chevron doesn’t open up dropdown (it also looks like a disabled color?)
> - Should clear out selected input from field if you open it again. (You should still see the check but a user should be able to type something else)
> - Chevrons aren’t matching up/diff sizes

WdfI66qg

## TODO

- [x] Input clearing behavior
- [x] Add `endContent` prop for a freeform content slot
- [x] Update `Combobox` and `Select` to use `endContent` for the chevron
- [x] Update ` Combobox.Category` and `Select.Category` to use same color and size icon

akdetrick:
closes #1044 

After speaking with Hillary, we decided to try using a clear icon instead of auto-clearing the input on focus. The auto-clearing behavior had some technical issues and could lead to a bad UX for a user tabbing through the form (imagine tabbing back to correct an error and creating another error by mistake when focus moves through the Combobox)

### TextInput and Input

- Add `startContent` and `endContent` for custom beginning and end icons/buttons in inputs
- Update the clear button and add an `onInputClear` callback

### DropdownTrigger, Select, and Combobox chevron icons

- Use `endContent` to create a darker chevron for dropdown indicators per design
- Update size and position of category summary chevrons to match

### Combobox

- Add clear icon that shows up when there is an input value. When the user clicks it, the downshift state is reset and the input is refocused for input.

## Screenshots
Screenshot 2023-10-04 at 11 38 55 AM
Screenshot 2023-10-04 at 11 33 57 AM
Screenshot 2023-10-03 at 11 38 43 PM
Screenshot 2023-10-04 at 11 41 54 AM

Oct-04-2023 16-29-23

### DIFF
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -141,7 +141,6 @@ const Combobox = ({
     Boolean(controlledInputValue)
   );
 
-  // TODO: state reducer
   const {
     isOpen,
     selectedItem,
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -137,7 +137,11 @@ const Combobox = ({
   }
 
   const [displayedItems, setDisplayedItems] = useState(items);
+  const [showClearButton, setShowClearButton] = useState(
+    Boolean(controlledInputValue)
+  );
 
+  // TODO: state reducer
   const {
     isOpen,
     selectedItem,
@@ -147,7 +151,6 @@ const Combobox = ({
     getItemProps,
     highlightedIndex,
     inputValue,
-    setInputValue,
     openMenu,
     reset,
   } = useCombobox({
@@ -180,6 +183,14 @@ const Combobox = ({
       }
       onChange(newSelection);
       onInputChange(newSelection);
+    },
+    onStateChange: ({ type }) => {
+      if (
+        type === useCombobox.stateChangeTypes.ItemClick ||
+        type === useCombobox.stateChangeTypes.InputChange
+      ) {
+        setShowClearButton(true);
+      }
     },
   });
 
@@ -296,11 +307,7 @@ const Combobox = ({
       // Reset filtered items every time user refocuses.
       // Subsequent changes in the input will re-filter the list.
       openMenu();
-      // If the user revisits the combobox after selecting an item,
-      // only show the selected item and clear the input. The list will
-      // be re-filtered as they type
       if (hasSelectedItem) {
-        setInputValue("");
         setDisplayedItems(items.filter(isSelectable));
       }
     }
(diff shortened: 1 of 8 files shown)


src/Combobox/index.js
```diff
@@ -137,7 +137,11 @@ const Combobox = ({
   }
 
   const [displayedItems, setDisplayedItems] = useState(items);
+  const [showClearButton, setShowClearButton] = useState(
+    Boolean(controlledInputValue)
+  );
 
+  // TODO: state reducer
```
akdetrick, line 144:
```suggestion
```

src/Combobox/index.js
```diff
@@ -327,20 +337,21 @@ const Combobox = ({
...
             onClick: handleMenuOpen,
-            onBlur: () => {
-              // If the user has selected an option, we should
-              // always set that as the value of the input.
-              if (hasSelectedItem) {
-                setInputValue(
-                  selectedItem.props.searchValue || selectedItem.props.value
-                );
-              }
-            },
```
akdetrick, line 343: This is now handled by the `onStateChange` downshift option

src/DropdownTrigger/index.js
```diff
@@ -48,13 +48,15 @@ const DropdownTrigger = React.forwardRef(
               {labelText}
             </label>
           )}
-          {displayValue && <span className="nds-dropdownTrigger-value">{displayValue}</span>}
+          {displayValue && (
+            <span className="nds-dropdownTrigger-value">{displayValue}</span>
+          )}
```
akdetrick, line 53: formatting only

src/TextInput/index.js
```diff
@@ -14,44 +14,61 @@ const TextInput = React.forwardRef((props, forwardedRef) => {
...
     ...nativeElementProps
   } = props;
 
+  const inputRef = useRef(null);
   const [inputValue, setInputValue] = useState(
     defaultValue ? defaultValue : ""
   );
 
+  // allows us to use the forwarded ref internally
+  useImperativeHandle(forwardedRef, () => inputRef.current);
```
akdetrick, line 37: This is my first time using this... it allows us to have access to the forwarded ref internally in this component so we may use it to set focus on the input when it's cleared

https://react.dev/reference/react/useImperativeHandle

src/Combobox/index.js
akdetrick, line 144:
```suggestion
```

akdetrick:
Changing this behavior again; new PR will be needed

PR #1062, opened 2023-10-25
fix(Tag): prevent label overflow

akdetrick:
Fixes issue where label text could overflow the Tag.

Now the tag will grow as large as the parent container allows, and truncate the label text with ellipsis after that.

Screenshot 2023-10-25 at 5 12 09 PM


kylekennedy6 (approved):
Thank you!

PR #1072, opened 2023-11-01
fix(row): set min-width on Row.Item

kevmo314:
This allows content to overflow if it sets `overflow` and has no effect if the content has an intrinsic width (most other cases).

For https://github.com/narmi/banking/issues/38767

image


akdetrick (approved):
This is an S-tier bug fix - Thanks!

`Row` and `whiteSpace--truncate` will finally play nicely together

PR #1083, opened 2023-11-13
feat(Alert): add Alert component

### LINKED ISSUE #1082: [Alert] Component for Inline toast-like system messages
## Short Description

Info, warn, and error boxes used to display "inline" system messages. For example, highlighting an error or update within a `ContentCard`. The `Alert` component is **not** modal, and lives in the document flow.

### Mockup

cc @hillarypollak 

Screenshot 2023-11-13 at 2 32 15 PM

### Responsibilities

- Dismiss callback
- Content props
- Allowing arbitrary JSX inside the content area
- `isOpen` prop. If the component controls this, we can ensure the Alert is in an aria-live region

- Timing (no auto dismiss)
- Dismissal (handled by consumer via `onUserDismiss` and `isOpen` prop)

### Props Signature

TK

```
<Alert
  propA={}
  propB={}
```

akdetrick:
closes #1082 

Adds `Alert` component for rendering system messages within document flow. Has an `aria-live` region built in.

Screenshot 2023-11-13 at 6 28 59 PM

### DIFF
--- a/src/Alert/index.jsx
+++ b/src/Alert/index.jsx
@@ -1,0 +1,73 @@
+import React from "react";
+import PropTypes from "prop-types";
+import cc from "classcat";
+import Row from "../Row";
+
+const noop = () => {};
+
+/**
+ * Inline system message, for a sepcific region of a page.
+ * The `isActive` prop is used to hide and show the Alert to ensure the Alert
+ * is always rendered in an [ARIA live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)
+ * for accessibility.
+ */
+const Alert = ({
+  isActive = false,
+  isDismissable = true,
+  onUserDismiss = noop,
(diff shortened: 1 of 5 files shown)


src/Alert/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+  isActive = false,
+  isDismissable = true,
+  onUserDismiss = noop,
+  kind = "info",
+  children,
+}) => {
+  const iconName = kind === "success" ? "check" : "info";
+
+  return (
+    <div aria-live="polite">
```
akdetrick, line 24: This is the live region. When you use an `Alert`, this div will _always_ get rendered. This is useful because assistive technology will announce the alert that renders inside it when `isActive` changes to `true`.

francesbhughes (approved):
Looks good!

akdetrick:
Awaiting design approval before merging

PR #1085, opened 2023-11-16
Feat/1081 new checkbox states

### LINKED ISSUE #1081: [Checkbox] Add indeterminate, disabled states; add size variant
### New Hover / Focus state
We want to render an alpha circle shim around the checkbox without it taking up any space in document flow. Use alpha to color mix with any background the checkbox is on.

### Disabled state
As `disabled` boolean prop.

### Indeterminate state
The NDS `Checkbox` should support an indeterminate state. This is **not** a variant, but [a state represented in the rendered DOM](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes) that should be represented by a controlled boolean prop, `indeterminate`.

### Sizes
This design introduces a small size variant for our checkbox component.

## Design

cc @jlnguyen91 

https://www.figma.com/file/aqnNhzzRMJvKrMcJiVrMd4/NDS-WIP?type=design&node-id=38-771&mode=design&t=dJ8Q1TGiwkTs6rE4-0

Screenshot 2023-11-13 at 2 18 14 PM
Screenshot 2023-11-13 at 2 18 06 PM

Screenshot 2023-11-13 at 2 39 05 PM

## Definition of Done

- Checkbox hover effect updated to circle for `table` type
- Checkbox supports `indeterminate` prop
- Checkbox has a new variant, `table`, that has a circle hover effect and is smaller
- Checkbox supports `disabled` prop

akdetrick:
closes #1081 

- Add indeterminate checkbox state
- Add `table` variant of checkbox with special hover state and hidden a11y label

### DIFF
--- a/src/Checkbox/index.js
+++ b/src/Checkbox/index.js
@@ -2,8 +2,8 @@
 import React, { useRef, useState, useEffect } from "react";
 import PropTypes from "prop-types";
 import cc from "classcat";
-import Error from "../Error";
 import ReactMarkdown from "react-markdown";
+import Error from "../Error";
 
 /**
  * Narmi styled checkbox with built-in label.
@@ -18,7 +18,6 @@ const Checkbox = ({
   checked,
   disabled = false,
   indeterminate = false,
-  size = "m",
   value,
   error,
   kind = "normal",
@@ -77,7 +76,6 @@ const Checkbox = ({
         className={cc([
           `nds-checkbox nds-checkbox--${kind}`,
           "fontWeight--default",
-          `nds-checkbox--${size}`,
           {
             "nds-checkbox--checked": isChecked || indeterminate,
             "nds-checkbox--disabled": disabled,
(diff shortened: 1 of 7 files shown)


akdetrick:
@jlnguyen91 @hillarypollak

Use the preview link posted in this thread to review Checkbox changes

### Open questions for design

1. It looks like the Figma designs set the normal checkbox at 20x20 (up from 18x18). I made the change in this branch, but are we sure we want to update that for every "normal" checkbox variant?
2. Should the circle hover effect apply to only the "table" variant of the checkbox, or should we update the "normal" variant to use that hover effect as well?

akdetrick:
Changes approved by design - marking ready for review

akdetrick:
@akdetrick follow up with design on priority and completeness of this PR

akdetrick:
Need to put up a new PR for this

PR #1090, opened 2023-11-24
fix(Dialog): prevent mobile chrome from overlapping dialog footer

### LINKED ISSUE #1087: [Dialog] Add extra padding on bottom when in responsive
Coming out of https://github.com/narmi/banking/issues/39451

The submit button is hidden on mobile web in the dialog so we probably want to add extra padding on the bottom to resolve this

akdetrick:
closes #1087 

CSS vocab:
```scss
property: value; // the pair is a "rule"
```

When a CSS ruleset has two different rules with the same property name, it will use the cascade (source order) to resolve the conflict. Browsers ignore rules that have properties or values they don't understand, so older browsers will just fall back to `100vh` in this case.

### DIFF
--- a/src/Dialog/index.scss
+++ b/src/Dialog/index.scss
@@ -46,7 +46,8 @@
 
   // the modal dialog takes over the whole screen by default
   // (small veiwports)
-  height: 100vh;
+  height: 100vh; // fall back to 100vh for older mobile browsers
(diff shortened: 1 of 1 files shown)


graphicsforge (approved):
today I learned about dvh!
👍

xiexiewenting (approved):
LGTM!

PR #1099, opened 2023-12-08
feat(Dialog): add notification content slot

### LINKED ISSUE #1098: [Dialog] add content container for pinned notification
Render inline `Alert` components in a special content slot in `Dialog` that ensures it's always visible.

akdetrick:
closes #1098 

Adds a pinned content zone to Dialog near the header that can accept an `Alert` or any other JSX content

### DIFF
--- a/src/Dialog/index.js
+++ b/src/Dialog/index.js
@@ -35,6 +35,7 @@ const Dialog = ({
   title,
   headerStyle = "bordered",
   children,
+  notification,
   footer,
   width = "500px",
   testId,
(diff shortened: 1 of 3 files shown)


kylekennedy6 (approved):
TY!!

graphicsforge (approved):
ah, a shame that the NDS doesn't have the NotificationContext, but I suppose we can move our dialog notification context to the outside of the dialog and then pass the notification down as a prop.

Thanks for taking a Node: the notification type can be handled by the notification context 🙏

Issue #456, opened 2021-12-07
[CI] Add license check via GitHub Actions to ensure dependencies use approved licenses


chris-griffin:
We should run a CI check on PRs to ensure that we do not introduce a dependency without an approved license

Resources:

https://medium.com/@fokusman/the-easiest-way-to-check-all-your-npm-dependency-licenses-753075ef1d9d
https://github.com/tangro/actions-license-check

reshav-abraham:
something of the sort works pretty well
`license-checker --onlyAllow 'MIT'; 'Apache-2.0'`

reshav-abraham:
Notes from Amy:
What is the timeline? Do we want this done by the end of the year?
What you’re looking for from a notice/checklist perspective?

akdetrick:
The scope of this ticket is to make sure we don't use any new vendors that don't comply with the licenses we already have in NDS.

For legal guidance, I filed a follow up ticket as that decision would apply to Narmi OSS practices more generally:
https://github.com/narmi/banking/issues/41618

PR #1100, opened 2023-12-11
feat(CI): check package dependencies for each pr

reshav-abraham:
for https://github.com/narmi/design_system/issues/456

Starting from the list 
https://opensource.org/licenses/?categories=popular-strong-community

However it seems like we have an unknown dependency


akdetrick (approved):

reshav-abraham:
We will the merge the ci check as is since that is the current list. We will open a separate issue to review the licenses with Amy

PR #1106, opened 2023-12-18
Ts convert/39969 compound components

akdetrick:
closes https://github.com/narmi/banking/issues/39969

Convert all compound components in NDS to typescript.

PropTypes are preserved for now, until I get storybook to build an ArgsTable from interfaces


src/Combobox/ComboboxCategory.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+interface ComboboxCategoryProps {
+  children: React.ReactNode;
+  label?: string;
+}
```
ejrb, line 6: React has a helper type PropsWithChildren you might be able to use for many of these
```ts
type ComboboxCategoryProps = React.PropsWithChildren<{
 label?: string;
}>;
```
the only caveat is `children` is optional, which you may not want in certain cases?

src/Combobox/index.tsx
```diff
@@ -115,19 +156,23 @@ const Combobox = ({
 }) => {
   const allChildren = React.Children.toArray(children);
```
ejrb, line 157: the type errors below are because children might not all be ReactElements - they could be string, or undefined, or other things which would cause errors.  you could tighten the props type to restrict just element-type nodes, or you could filter this `allChildren` using a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) function

```ts
const hasDisplayName = (e: any): e is React.ReactElement =>
  !!e?.type?.displayName;

const allChildren = React.Children.toArray(children).filter(hasDisplayName);
```

then allChildren would be compatible with the filtering below

src/Combobox/ComboboxItem.tsx
```diff
@@ -1,7 +1,27 @@
 import React from "react";
 import PropTypes from "prop-types";
```
ejrb, line 2: what do you think about removing prop-types from components converted to TS?

src/MenuButton/MenuButtonItem.tsx
```diff
@@ -1,13 +1,26 @@
...
-import iconSelection from "src/icons/selection.json";
+import iconSelection from "../icons/selection.json";
 import Row from "../Row";
+import { IconName } from "../types/Icon.types";
+
+interface MenuButtonItemProps {
+  /** Display label for menu item */
+  label: string;
+  /** Selection handler */
+  onSelect: () => void;
```
ejrb, line 11: `onSelect` not used?

edit: oh I see, it's for the parent to pull from the props?  perhaps a future refactor could group things with this handle

```ts
export interface SelectableChildProps {
   /** Selection handler */
  onSelect: () => void;
}

interface MenuButtonItemProps extends SelectableChildProps {
   ...
}
```

ejrb (commented):
Looks great.  It feels like TS and prop-types are fighting each other a bit.  Is prop-types still useful for runtime validation, or could we start to remove it?

src/Combobox/ComboboxCategory.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+const ComboboxCategory: React.FC<ComboboxCategoryProps> = ({ children }) => (
```
kevmo314, line 8: `React.FC` is less popular these days: https://github.com/facebook/create-react-app/pull/8177

```
const ComboboxCategory = ({ children }: React.PropsWithChildren<{}>) => ...
```

for the alternative.

src/IconButton/index.js
```diff
@@ -22,7 +22,7 @@ const IconButton = ({
   testId = "nds-icon-button",
   label,
   name,
-  onClick = () => {},
+  onClick = (e) => {}, // eslint-disable-line no-unused-vars
```
kevmo314, line 25:
```suggestion
  onClick = (_) => {},
```

src/Select/index.tsx
```diff
@@ -96,7 +136,7 @@ export const isSelectedItemInCategory = (selectedItem, categoryChildren) => {
  * @param {String} userInput most recent thing a user typed while focused on input
  * @returns {String} the string to use for typeahead for each given `selectItem`
  */
-// eslint-disable-next-line no-unused-vars
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const defaultGetTypeAheadString = (userInput = "", selectItem) => {
```
kevmo314, line 140: I'm surprised it doesn't complain about this, it seems odd that a non-optional parameter comes after an optional one? Are the items in the wrong order, the docs suggest it should be `selectItem` first?

src/Combobox/ComboboxItem.tsx
akdetrick: I agree, but I want to do that in a separate step to ensure we don't lose the args tables in storybook documentation as we convert over

src/IconButton/index.js
akdetrick: Lovely - should have thought about that!

src/Combobox/index.tsx
akdetrick: Wow, this is pretty cool

akdetrick:
Back to draft until all feedback can be addressed

akdetrick:
Putting up a new PR without `React.FC`

PR #1107, opened 2024-01-02
feat(CollapsibleCard): add CollapsibleCard component

rogerioshieh:
Adds a `CollapsibleContent` component:

https://github.com/narmi/design_system/assets/40327446/f17c0489-811e-4d86-9e5a-fa0d84407ee9


src/CollapsibleCard/index.js
```diff
@@ -0,0 +1,157 @@
...
+  children: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.arrayOf(PropTypes.node),
+  ]).isRequired,
+  /** Card title */
+  title: PropTypes.string,
+  /** Card subtitle */
+  subtitle: PropTypes.string,
+  /** Card helper text, placed on the right side of the title container */
+  helperText: PropTypes.node,
```
akdetrick, line 140: **Question**

Would it make sense to call this `statusText`? I think it works either way

src/CollapsibleCard/index.js
```diff
@@ -0,0 +1,157 @@
...
+};
+
+CollapsibleCard.propTypes = {
+  /** Accepts any content as children */
+  children: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.arrayOf(PropTypes.node),
+  ]).isRequired,
+  /** Card title */
+  title: PropTypes.string,
```
akdetrick, line 136: **Suggestion**

This prop should probably be required

src/CollapsibleCard/index.js
```diff
@@ -0,0 +1,157 @@
...
+  /** Card subtitle */
+  subtitle: PropTypes.string,
+  /** Card helper text, placed on the right side of the title container */
+  helperText: PropTypes.node,
+  /** Controls whether card is opened */
+  isOpen: PropTypes.bool.isRequired,
```
akdetrick, line 142: **Suggestion**

The `isOpen` prop is already defaulting to `false` in the component definition, so we could make this optional (also, undefined is falsy)


```
const CollapsibleCard = ({
  title,
  subtitle,
  helperText,
  isOpen = false, // <------------
  onOpen = () => {},
  onClose = () => {},
  isDisabled = false,
  onDisabledClick = () => {},
  hasError = false,
  disableHover = false,
  children,
}) => {
```

src/CollapsibleCard/index.js
```diff
@@ -0,0 +1,157 @@
...
+  /** Callback to handle user opening card */
+  onOpen: PropTypes.func,
+  /** Callback to handle user closing card */
+  onClose: PropTypes.func,
```
akdetrick, line 146: **Question/Suggestion**

What if we only had one `onClick` callback that was called as `(isOpen) => {}`?

Example:

```jsx
<CollapsibleCard
    onClick={(isOpen) => { console.log(`The card is now ${isOpen ? 'open' : 'closed'}`) }}
/>
```

```jsx
<CollapsibleCard
    isOpen={isOpenFromParentState}
    onClick={(isOpen) => {
        setIsOpenFromParentState(isOpen)
    }}
```

src/CollapsibleCard/index.js
```diff
@@ -0,0 +1,157 @@
...
+  /** Disabled cards are greyed out and do not open */
+  isDisabled: PropTypes.bool,
+  /** Callback to handle user clickling on disabled card */
+  onDisabledClick: PropTypes.func,
```
akdetrick, line 150: **Question**

Do we know of any use cases where we need this callback? We could remove this prop for now and put it back only if we need it. If we refactor `onOpen`/`onClose` into a single `onClick` callback, we could simply not call `onClick` when the card is disabled.

akdetrick (commented):
left some questions

akdetrick:
**Optional/Minor**

Screenshot 2024-01-09 at 2 21 10 PM

Sometimes a click+drag when opening the card will select all the text. This doesn't bother me, but it may bother design a bit.

You might be able to add `user-select: none` to the title

src/CollapsibleCard/index.js
rogerioshieh: Agree, `statusText` is a better name 💯

src/CollapsibleCard/index.js
rogerioshieh: The CollapsibleContentCard in cerulean (which this is based on) actually has different callbacks for onOpen and onClose. For forms, for instance (which is how we're currently using it in cerulean), often we will want to present a toast error or run a validation onClose, but not onOpen, so having separate callbacks would allow for more flexibility.

src/CollapsibleCard/index.js
rogerioshieh: Yes--currently, the way we're using it is to show a toast notification when you try to open a disabled card if this prop is passed. This is relevant for a page with multiple cards where one card is only enabled after a specific card is already completed (which is how we currently use this).

akdetrick (approved):

PR #1110, opened 2024-01-09
Refactor/1103 reachui to react aria

### LINKED ISSUE #1103: [dependencies] Reach-UI blocks upgrade to React 18
via support:

> When we try to upgrade to the latest version of node we are getting an dependency error. The narmi/design_system package does claim to support 18.2 but the @reach.menu-button does not support 18. So we cannot upgrade.
See the dependency map below:
"dependencies": {
 "@reach/menu-button": "^0.18.0",
 "classcat": "^5.0.3",
 "downshift": "^6.1.7",
 "eslint-plugin-json": "^3.1.0",
 "flatpickr": "^4.6.12",
 "raf-schd": "^4.0.3",
 "react-focus-lock": "^2.9.4",
 "react-laag": "^2.0.3",
 "react-markdown": "^6.0.3",
 "react-transition-group": "^4.4.5",
 "usehooks-ts": "^2.9.1"
 },

akdetrick:
closes #1103 

- Refactors `MenuButton` to use `react-aria` library instead of `@reach`.
- Removes `@reach` dependency

**No breaking changes.** The props signature and behaviors will be the same as before.

react-aria docs: https://react-spectrum.adobe.com/react-aria/Menu.html

This does increase the JS payload for NDS but a bit, but it's worth it in the long run. The `react-aria` library would be my recommended base for components moving forward, and we also have the opportunity to remove the `downshift` dependency.

### DIFF
--- a/src/MenuButton/index.scss
+++ b/src/MenuButton/index.scss
@@ -22,54 +22,27 @@
   }
 }
 
-/**
-* Reach MenuButton theme
-*/
-:root {
-  --reach-menu-button: 1;
-}
-
-[data-reach-menu] {
-  position: relative;
-}
-
-[data-reach-menu-popover] {
-  display: block;
-  position: absolute;
-}
-
-[data-reach-menu-popover][hidden] {
-  display: none;
-}
-
-[data-reach-menu-list],
-[data-reach-menu-items] {
-  display: block;
+.nds-menubutton-menu {
   white-space: nowrap;
   box-shadow: var(--elevation-high);
   outline: none;
 }
 
-[data-reach-menu-item] {
-  display: block;
+.nds-menubutton-item {
   user-select: none;
-}
-
-[data-reach-menu-item] {
   cursor: pointer;
   display: block;
   color: inherit;
   font: inherit;
   text-decoration: initial;
-}
-
-/* pseudo pseudo selector */
-[data-reach-menu-item][data-selected] {
-  background: rgba(var(--theme-rgb-primary), var(--alpha-5));
   outline: none;
-}
 
-[data-reach-menu-item][aria-disabled] {
-  opacity: 0.5;
-  cursor: not-allowed;
+  &--highlighted {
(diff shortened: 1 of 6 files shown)


tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
     "outFile": "./dist/types/index",
-    "target": "es5",
+    "target": "ES6",
+    "lib": ["ES2022", "dom"],
```
akdetrick, line 5: needed for react-aria usage of `Intl`

src/MenuButton/index.stories.js
```diff
@@ -64,7 +64,13 @@ export const InADialog = () => {
       >
         Open Dialog
       </button>
-      <Dialog isOpen={isOpen} title="Dialog with a MenuButton">
+      <Dialog
+        isOpen={isOpen}
+        title="Dialog with a MenuButton"
+        onUserDismiss={() => {
+          setIsOpen(false);
+        }}
```
akdetrick, line 72: Fixed an existing bug where the dialog in this story wouldn't close

src/MenuButton/index.scss
```diff
@@ -22,54 +22,32 @@
   }
 }
 
-/**
-* Reach MenuButton theme
-*/
-:root {
-  --reach-menu-button: 1;
+.nds-menubutton-popover {
+  z-index: 4;
```
JohnathanWeisner, line 26: z-indexes scare me

JohnathanWeisner:
Hover styling is off compared to the current version.

src/MenuButton/index.scss
akdetrick: So everything that portals in NDS portals to `document.body`, so the order in which they get created helps set the stacking context.

Here, I'm setting 4 because the Dialog is 3

JohnathanWeisner (approved):
Thanks for the fixes! Looks great.

PR #1114, opened 2024-01-18
feat(typography): use real weight and italics instead of faux bolding…

akdetrick:
closes https://github.com/narmi/banking/issues/40039

We have been using faux italics in body type. This PR ensures the italic font variation setting from the variable font is enabled, and uses the `font-synthesis` property to prevent faux bolding and faux slanting.

lovely italic glyphs as the type designer intended (good):
Screenshot 2024-01-17 at 7 00 16 PM

the browser just applying a slant transform (bad):
Screenshot 2024-01-17 at 7 00 22 PM

### DIFF
--- a/src/base-styles/defaults.scss
+++ b/src/base-styles/defaults.scss
@@ -15,6 +15,7 @@ body {
   max-width: unset;
   background-color: var(--color-white);
   text-decoration-skip-ink: none;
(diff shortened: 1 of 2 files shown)


graphicsforge (approved):
Huh, I didn't know the browser could do that either!

PR #1116, opened 2024-01-18
fix(MenuButton): prevent popover from closing before selection callback

akdetrick:
Fixes this issue reported by @JohnathanWeisner 

edit-wire-recipient-not-opening (1)

### DIFF
--- a/src/MenuButton/index.js
+++ b/src/MenuButton/index.js
@@ -72,12 +72,22 @@ const MenuButton = ({
       (item) => labelToItemId(item.props.label) === itemId
     );
     selectedItem.props.onSelect();
+    closePopover();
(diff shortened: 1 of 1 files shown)


src/MenuButton/index.js
```diff
@@ -72,12 +72,22 @@ const MenuButton = ({
       (item) => labelToItemId(item.props.label) === itemId
     );
     selectedItem.props.onSelect();
+    closePopover();
```
akdetrick, line 75: we want the popover to close after selection

JohnathanWeisner (approved):
Looks good. One follow up would be that you can't close the menu by clicking it again:

PR #1118, opened 2024-01-18
docs(TextInput): add example of hide/show password

### LINKED ISSUE #824: [TextInput]: Add story with example for password TextInput w/ Show button
The specific use case is to have a "Show" button inside of TextInput that will reveal the characters entered when the input it hidden

mostly, this is useful for a password input

Figma and screenshot included below:
Screen Shot 2022-10-13 at 3 01 31 PM

https://www.figma.com/file/YWsfmcvcQMak3hzxA0yNwV/enrollments?node-id=2%3A3788

akdetrick:
closes #824 

Adds storybook story to show how to build a password input with show/hide feature

Screenshot 2024-01-18 at 5 41 43 PM
Screenshot 2024-01-18 at 5 41 50 PM

### DIFF
--- a/src/TextInput/index.stories.js
+++ b/src/TextInput/index.stories.js
@@ -1,5 +1,6 @@
 import React, { useState } from "react";
 import TextInput, { VALID_ICON_NAMES } from "./";
+import Button from "../Button";
 
 const Template = (args) => <TextInput {...args} />;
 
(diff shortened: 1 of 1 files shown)


kylekennedy6 (approved):
LGTM! @graphicsforge I think we may want to revisit our implementation of `PasswordTextInput` in cerulean, since it seems like we could leverage the `endContent` prop rather than what we're currently doing: https://github.com/narmi/banking/blob/cf39deb78eeea078c90f5f77b6872552f08bf3dc/cerulean/src/PasswordTextInput/index.tsx#L5

PR #1119, opened 2024-01-18
feat(ProgressBar): add ProgressBar component

### LINKED ISSUE #1104: [Progress] add Progress component
### Short Description

We have a progress bar a few places in Narmi apps and we generally do it by changing the width of an element with a background.

We could do this with SVG instead, and it would be nice to have one configurable progress bar instead of reimplementing it every time

### Mockup

see: https://github.com/narmi/banking/pull/40315

### Responsibilities

- Taking a % complete number
- Configuring total width
- Animating the progress change when props change

- calculating actual progress (it must be passed in)
- anything outside of the progress bar itself - it needs to be flexible enough to be placed anywhere

### Props Signature

```
<Progress
  percentComplete={Number}
  width={Number (pixels)}
```

akdetrick:
closes #1104 

Adds `ProgressBar`.

We can add a `kind` prop later if we need to, but for now, we match the style of the file upload progress bar.

### DIFF
--- a/src/ProgressBar/index.js
+++ b/src/ProgressBar/index.js
@@ -1,0 +1,34 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const ProgressBar = ({ totalWidth = 120, percentComplete = 0 }) => {
+  return (
+    <div
+      className="nds-progressbar"
+      style={{ width: `${totalWidth}px`, height: "18px" }}
+    >
+      <svg xmlns="http://www.w3.org/2000/svg">
+        <line
+          x1="0%"
+          y1="50%"
+          x2={`${percentComplete}%`}
+          y2="50%"
+          strokeWidth="100%"
+        />
+      </svg>
+    </div>
+  );
+};
+
+ProgressBar.propTypes = {
+  /**
+   * Total width of progress bar
+   */
+  totalWidth: PropTypes.number,
+  /**
+   * Percent complete
+   */
+  percentComplete: PropTypes.number.isRequired,
+};
+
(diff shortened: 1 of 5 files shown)


src/ProgressBar/index.scss
```diff
@@ -0,0 +1,15 @@
+.nds-progressbar {
+  border: 1px solid var(--theme-primary);
+  border-radius: var(--border-radius-default);
+  //overflow: hidden;
```
akdetrick, line 4:
```suggestion
```

chris-griffin:
Looks great! Only thing that came to mind is that it might be more helpful to be able to pass in the width with the units (e.g. `120px` or `50%`), but no strong feelings about it.

src/ProgressBar/index.js
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const ProgressBar = ({ totalWidth = "100%", percentComplete = 0 }) => {
```
akdetrick, line 4: @chris-griffin Thanks for taking a look - this was a good catch!

I kind of broke my own rules here, because the best practice would be to make components fill the width of their parent container, unless otherwise specified by props. We can default this to `100%` and allow consumers to pass in a complete CSS value with the unit

chris-griffin (approved):
Very clean implementation, thanks!

I'm assuming design will always want the progress bar to be 18px in height. If that is not the case, it might be worth adding the prop now.

Whenever this passes a design review, it looks good to merge to me.

PR #1121, opened 2024-02-02
chore(Alert): add data-testid for the dismiss button

graphicsforge:
Just adding a test id so that we can tell if there is a dismiss button or not in tests, and to test clicking it!


kylekennedy6 (approved):
LGTM unless @akdetrick has any reservations!

JohnathanWeisner (approved):
LGTM 2

PR #1123, opened 2024-02-07
feat(checkbox): add indeterminate state property

kevmo314:
Permits checkboxes to be rendered as indeterminate by passing `indeterminate`.

image

### DIFF
--- a/src/Checkbox/index.stories.js
+++ b/src/Checkbox/index.stories.js
@@ -43,7 +43,7 @@ export const Indeterminate = () => {
   return (
     <>
       <Checkbox
-        label="Intermediate state"
+        label="Indeterminate state"
         indeterminate
         checked={isChecked}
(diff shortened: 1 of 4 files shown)


src/Checkbox/index.stories.js
```diff
@@ -38,6 +38,33 @@ export const MultipleCheckboxes = () => (
   </>
 );
 
+export const Indeterminate = () => {
+  const [isChecked, setIsChecked] = useState(false);
+  return (
+    <>
+      <Checkbox
+        label="Intermediate state"
```
elliot-wilson, line 46: is this a typo? these words sound very similar

elliot-wilson (approved):
I tried it out in the Storybook Preview. Seems like it works! What's the purpose?

src/Checkbox/index.stories.js
kevmo314: Nice catch, thanks!

PR #1124, opened 2024-02-12
feat(popover): allow popover to be controlled

kevmo314:
Allows `Popover` to accept `isOpen` and `onUserDismiss`, following the same pattern as `Dialog` and `Tooltip`.

### DIFF
--- a/src/Popover/index.js
+++ b/src/Popover/index.js
@@ -3,6 +3,7 @@ import React, { useState, useEffect } from "react";
 import PropTypes from "prop-types";
 import { useLayer } from "react-laag";
 import FocusLock from "react-focus-lock";
+import noop from "lodash";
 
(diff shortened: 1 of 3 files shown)


src/Popover/index.js
```diff
@@ -26,41 +27,48 @@ const Popover = ({
   matchTriggerWidth = false,
   testId,
   closeOnContentClick = false,
+  isOpen,
+  onUserDismiss = noop,
 }) => {
-  const [isOpen, setIsOpen] = useState(false);
+  const isControlled = isOpen === true || isOpen === false;
```
elliot-wilson, line 33: neat!

elliot-wilson (approved):
LGTM! I tested with the storybook

PR #1127, opened 2024-02-13
feat(alert): Customizable icon and new alert kind

aundus:
New designs for fednow require additional customization of the alert component. 

Designs: https://www.figma.com/file/TG6rH21v231JFcZy9jaIz6/fednow-manager?type=design&node-id=1101-2&mode=design&t=942hC19EsHzktDbo-0

Dev storybook:
image

### DIFF
--- a/src/Alert/index.jsx
+++ b/src/Alert/index.jsx
@@ -64,7 +64,7 @@ Alert.propTypes = {
   /** Callback for user dismissal actions */
   onUserDismiss: PropTypes.func,
   /** Variant of Alert to use */
-  kind: PropTypes.oneOf(["info", "error", "success", "plain"]),
+  kind: PropTypes.oneOf(["info", "error", "success", "warn"]),
   /** Override the default icon of the alert */
   icon: PropTypes.string,
   /** Message content of the Alert */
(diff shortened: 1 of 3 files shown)


src/Alert/index.jsx
```diff
@@ -63,7 +64,9 @@ Alert.propTypes = {
   /** Callback for user dismissal actions */
   onUserDismiss: PropTypes.func,
   /** Variant of Alert to use */
-  kind: PropTypes.oneOf(["info", "error", "success"]),
+  kind: PropTypes.oneOf(["info", "error", "success", "plain"]),
```
akdetrick, line 67: Based on conversations with design, this new type of alert should be `info`, and the idea was to add `warn` to have the styles of the current `info` Alert.

```suggestion
  kind: PropTypes.oneOf(["info", "error", "success", "warn"]),
```

akdetrick (approved):
annoying but necessary change for `Alert` moving forward

src/Alert/index.jsx
aundus: Yeah - I was thinking that naming would make more sense, but didnt want to change the current naming convention to break anything. I'll look around the repo when I bump the NDS version to make sure that we change the kind of any alerts that need it.

PR #1129, opened 2024-02-14
feat(drawer): allow children to be a render function

kevmo314:
Adds support for a render function children in `Drawer`. This allows content inside the drawer to be rendered on-demand and destroyed when the drawer is closed. Notably, such behavior cannot be accomplished by conditioning on `isOpen` because `isOpen` would destroy the contents too quickly and we want to wait until the drawer is fully closed.

See storybook example for a more descriptive demonstration.

### DIFF
--- a/src/Drawer/index.stories.js
+++ b/src/Drawer/index.stories.js
@@ -259,7 +259,7 @@ LazyLoadedContent.parameters = {
   docs: {
     description: {
       story:
-        "The Drawer component accepts a render prop for its children. This can be used to lazy load content or to render content conditionally.",
+        "The Drawer component accepts a render prop for its children with argument `isVisible`. This can be used to lazy load content or to render content conditionally.",
     },
   },
 };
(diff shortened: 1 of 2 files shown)


src/Drawer/index.stories.js
```diff
@@ -230,6 +230,40 @@ export const ContentWithDialog = () => {
...
+        }}
+      </Drawer>
+    </>
+  );
+};
+LazyLoadedContent.parameters = {
+  docs: {
+    description: {
+      story:
+        "The Drawer component accepts a render prop for its children. This can be used to lazy load content or to render content conditionally.",
```
elliot-wilson, line 262: Would it be helpful to explicitly state that the render prop will receive `isVisible`?

elliot-wilson (approved):
Makes sense, and now I know more about [render props](https://legacy.reactjs.org/docs/render-props.html#using-props-other-than-render)!

PR #1145, opened 2024-03-11
feat(IconButton): allow passing type to IconButton

elliot-wilson:
I'd like to allow IconButtons to control whether the underlying `<button>` is of type "button" vs "submit" vs "reset".

### DIFF
--- a/src/IconButton/index.js
+++ b/src/IconButton/index.js
@@ -12,8 +12,6 @@ export const VALID_ICON_NAMES = iconSelection.icons.map(
  *
  * Icon Button renders a button with a central Narmi icon
  *
- * This component supports rest props; any additional props on icon button will be
- * passed through to button element.
  */
 const IconButton = ({
   disabled = false,
@@ -22,10 +20,12 @@ const IconButton = ({
   testId = "nds-icon-button",
   label,
   name,
-  onClick = () => {},
+  onClick = () => { },
+  type,
 }) => {
   return (
     <button
+      type={type}
       onClick={onClick}
       className={cc([
         "nds-typography",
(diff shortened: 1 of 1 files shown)


src/IconButton/index.js
```diff
@@ -12,8 +12,6 @@ export const VALID_ICON_NAMES = iconSelection.icons.map(
  *
  * Icon Button renders a button with a central Narmi icon
  *
- * This component supports rest props; any additional props on icon button will be
- * passed through to button element.
```
elliot-wilson, line 16: removing - not actually true: see https://github.com/narmi/design_system/pull/940#discussion_r1231586350

src/IconButton/index.js
```diff
@@ -54,6 +54,8 @@ IconButton.propTypes = {
   disabled: PropTypes.bool,
   /** style of icon button to render */
   kind: PropTypes.oneOf(["action", "plain"]),
+  /** type attribute for underlying HTML button element */
+  type: PropTypes.string,
```
elliot-wilson, line 58: I could be pressured into making this `PropTypes.oneOf(["button", "submit", "reset"])`. My gut instinct was that, since the valid values are controlled by the HTML gods, not us, I'd rather not be too specific.

kevmo314 (approved):

src/IconButton/index.js
```diff
@@ -22,10 +20,12 @@ const IconButton = ({
   testId = "nds-icon-button",
   label,
   name,
-  onClick = () => {},
+  onClick = () => { },
+  type,
```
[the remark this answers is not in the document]
elliot-wilson: I'm okay with leaving it undefined -- and therefore defaulting to the browser default -- for now

PR #1152, opened 2024-03-19
fix(Popover): don't bubble up the user events that we handle

graphicsforge:
resolves https://github.com/narmi/banking/issues/43694

If a Popover is inside another component, like a tab or a summary, interacting with the Popover will also interact with the parent. Calling a stopPropagation() on the events that we handle in this component should prevent this bubbling.


akdetrick (approved):
Thanks!

PR #1156, opened 2024-03-22
feat(Drawer): use takeover drawer for mobile sized screens

### LINKED ISSUE #1132: [Drawer] make component responsive for mobile viewports
- Controls should move inside the Drawer
- Drawer should take over full viewport

kylekennedy6:
closes #1132

- For mobile screen sizes, renders a full screen width drawer & moves controls to the top of the drawer

### DIFF
--- a/src/Drawer/index.js
+++ b/src/Drawer/index.js
@@ -38,10 +38,7 @@ const Drawer = ({
 
   // The depth is how far the drawer opens out, but the CSS prop depends
   // on whether the Drawer is vertical or not
-  const depthStyle = {
-    width: isHorizontal ? "auto" : depth,
-    height: isHorizontal ? depth : "100%",
-  };
+  const depthStyle = isHorizontal ? { height: depth } : { width: depth };
   if (isVerticalMobileDrawer) {
     depthStyle.width = "100%";
   }
--- a/src/Drawer/index.js
+++ b/src/Drawer/index.js
@@ -38,7 +38,14 @@ const Drawer = ({
 
   // The depth is how far the drawer opens out, but the CSS prop depends
   // on whether the Drawer is vertical or not
-  const depthStyle = isHorizontal ? { height: depth } : { width: !isVerticalMobileDrawer ? depth : "100%" };
+  const depthStyle = {
+    width: isHorizontal ? "auto" : depth,
+    height: isHorizontal ? depth : "100%",
+  };
+  if (isVerticalMobileDrawer) {
+    depthStyle.width = "100%";
+  }
+
   useLockBodyScroll(isOpen);
 
   const handleKeyDown = ({ key }) => {
@@ -155,9 +162,14 @@ const Drawer = ({
   const childrenJSX = (
     <div
       style={depthStyle}
-      className={`drawer drawer-children drawer--${position} ${
-        isOpen && isTransitioning ? `drawer--open--${position}` : ""
-      } ${isVerticalMobileDrawer ? "drawer--vertical--mobile" : ""}`}
+      className={cc([
+        "drawer",
+        `drawer--${position}`,
+        {
+          [`drawer--open--${position}`]: isOpen && isTransitioning,
+          "drawer--vertical--mobile": isVerticalMobileDrawer,
+        },
+      ])}
       role="dialog"
       data-testid={testId}
     >
--- a/src/Drawer/index.js
+++ b/src/Drawer/index.js
@@ -172,20 +172,24 @@ const Drawer = ({
     >
       {isVerticalMobileDrawer && (
         <>
-          <div
-            onClick={onPrev}
-          >
-            <span
-              className="narmi-icon-chevron-left fontSize--heading3"
-            />
-          </div>
-          <div
-            onClick={onNext}
-          >
-            <span
-              className="narmi-icon-chevron-right fontSize--heading3"
-            />
-          </div>
+          {showControls && (
+            <>
+              <div
+                onClick={onPrev}
+              >
+                <span
+                  className="narmi-icon-chevron-left fontSize--heading3"
+                />
+              </div>
+              <div
+                onClick={onNext}
+              >
+                <span
+                  className="narmi-icon-chevron-right fontSize--heading3"
+                />
+              </div>
+            </>
+          )}
           <div
             onClick={onUserDismiss}
           >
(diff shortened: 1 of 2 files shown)


src/Drawer/index.js
```diff
@@ -153,13 +157,23 @@ const Drawer = ({
       style={depthStyle}
       className={`drawer drawer--${position} ${
         isOpen && isTransitioning ? `drawer--open--${position}` : ""
-      }`}
+      } ${isUsingMobileDrawer ? "drawer--mobile" : ""}`}
```
akdetrick, line 160: `import cc from "classcat"` will help here - a lot of components in NDS already use it to compose conditional classnames

akdetrick:
```jsx
className={cc([
    "thisClassIsAlwaysApplied",
    "soIsThisOne",
    {
        "conditionallyAddedClass": true, // will be added
        "AnotherconditionallyAddedClass": false, // will NOT be added
    }
])}
```

src/Drawer/index.js
```diff
@@ -153,13 +157,22 @@ const Drawer = ({
...
       data-testid={testId}
     >
-      {typeof children === "function"
-        ? children({ isVisible: isTransitioning })
-        : children}
+      {isVerticalMobileDrawer && (
+        <div
+          onClick={onUserDismiss}
+        >
+          <span className="narmi-icon-x clickable fontSize--heading3" />
```
kylekennedy6, line 168: TODO: we will want to add the left/right buttons here as well

src/Drawer/index.js
```diff
@@ -153,13 +157,22 @@ const Drawer = ({
       style={depthStyle}
       className={`drawer drawer--${position} ${
         isOpen && isTransitioning ? `drawer--open--${position}` : ""
-      }`}
+      } ${isVerticalMobileDrawer ? "drawer--vertical--mobile" : ""}`}
```
akdetrick, line 160: I would recommend `import cc from "classcat"` and use `cc([])` to build the class name. 

https://www.npmjs.com/package/classcat

src/Drawer/index.js
```diff
@@ -31,10 +32,13 @@ const Drawer = ({
...
 
   const isTransitioning = useMountTransition(isOpen, 300);
+  const { m } = useBreakpoints();
   const isHorizontal = position === "bottom" || position === "top";
+  const isVerticalMobileDrawer = (!m) && !isHorizontal;
+
   // The depth is how far the drawer opens out, but the CSS prop depends
   // on whether the Drawer is vertical or not
-  const depthStyle = isHorizontal ? { height: depth } : { width: depth };
+  const depthStyle = isHorizontal ? { height: depth } : { width: !isVerticalMobileDrawer ? depth : "100%" };
```
akdetrick, line 41: We may be able to simplify this a bit:

```jsx
const depthStyle = {
   width: isHorizontal ? "auto" : depth,
   height: isHorizontal ? depth : "auto",
}

if (isVerticalMobileDrawer) {
   depthStyle.width = "100%";
}
```

src/Drawer/index.js
```diff
@@ -151,12 +162,19 @@ const Drawer = ({
   const childrenJSX = (
     <div
       style={depthStyle}
-      className={`drawer drawer--${position} ${
+      className={`drawer drawer-children drawer--${position} ${
```
kylekennedy6, line 165: added the `drawer-children` class so we could distinguish the padding applied to the `.drawer.navigation` from the padding applied this container

src/Drawer/index.scss
```diff
@@ -12,7 +12,6 @@
   transition: transform var(--transition-speed-fast) ease;
   border: var(--border-size-s);
   border-color: var(--border-color-light);
-  padding: var(--space-xxl);
```
kylekennedy6, line 15: my thinking here was that we always want to apply this amount of padding to the `.drawer.navigation` container. But for the actual drawer contents (the children we're passing in), we want to apply a minimum padding regardless of whether the drawer is opening from top/bottom or left/right

src/Drawer/index.js
```diff
@@ -31,10 +32,21 @@ const Drawer = ({
...
   // The depth is how far the drawer opens out, but the CSS prop depends
   // on whether the Drawer is vertical or not
-  const depthStyle = isHorizontal ? { height: depth } : { width: depth };
+  const depthStyle = {
+    width: isHorizontal ? "auto" : depth,
+    height: isHorizontal ? depth : "auto",
+  };
+  if (isVerticalMobileDrawer) {
+    depthStyle.width = "100%";
+    depthStyle.height = "100%";
```
kylekennedy6, line 47: needed to also set height to be 100% here - otherwise it was only occupying a small amount of the vertical space

src/Drawer/index.js
```diff
@@ -31,10 +32,20 @@ const Drawer = ({
...
+  const { m } = useBreakpoints();
   const isHorizontal = position === "bottom" || position === "top";
+  const isVerticalMobileDrawer = (!m) && !isHorizontal;
+
   // The depth is how far the drawer opens out, but the CSS prop depends
   // on whether the Drawer is vertical or not
-  const depthStyle = isHorizontal ? { height: depth } : { width: depth };
+  const depthStyle = {
+    width: isHorizontal ? "auto" : depth,
+    height: isHorizontal ? depth : "100%",
```
kylekennedy6, line 43: I _think_ this need to be `100%` - with `auto`, the drawer was doing this
Screenshot 2024-03-28 at 1 34 23 PM

akdetrick: I don't doubt I did something wrong here...

Would this do it?

```
const depthStyle = isHorizontal ? { height: depth } : { width: depth };

if (isVerticalMobileDrawer) {
   depthStyle.width = "100%";
}
```

kylekennedy6: > I don't doubt I did something wrong here...
> 
> Would this do it?
> 
> ```
> const depthStyle = isHorizontal ? { height: depth } : { width: depth };
> 
> if (isVerticalMobileDrawer) {
>    depthStyle.width = "100%";
> }
> ```
> 
> I'm avoiding a nested ternary. We get a reasonable default style object, but if the mobile condition is true, we always set width to 100

Appears to have done the trick!

akdetrick (approved):

Issue #1153, opened 2024-03-21
[RadioButtons] Add details for card RadioButtons


JohnathanWeisner:
Add details for RadioButtons options.

PR #1163, opened 2024-03-29
feat(RadioButtons): Add details to the RadioButtons component

JohnathanWeisner:
resolves: #1153

This update adds details to the options property for the `RadioButtons` component.

### Normal radio buttons with details always shown:
Screenshot 2024-03-29 at 12 31 08 PM

### DIFF
--- a/src/RadioButtons/index.stories.js
+++ b/src/RadioButtons/index.stories.js
@@ -84,6 +84,53 @@ FullyControlled.parameters = {
   },
 };
 
+export const AsRadioButtonsWithDetails = Template.bind({});
+AsRadioButtonsWithDetails.args = {
+  options: {
+    OptionA: {
+      value: "A",
+      details:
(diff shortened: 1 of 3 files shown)


src/RadioButtons/index.stories.js
```diff
@@ -84,6 +84,53 @@ FullyControlled.parameters = {
...
+    OptionA: {
+      value: "A",
+      details:
+        "Option A details. The Option A details are very long and will wrap to the next line. The Option A details are very long and will wrap to the next line.",
+    },
+    OptionB: { value: "B", details: "Option B details" },
+    OptionC: { value: "C", details: "Option C details" },
+  },
+  name: "card_options_with_details",
+};
```
akdetrick, line 99: Is this a breaking change for existing RadioButtons usage?

If so, maybe we could set up a separate `optionsDetails` prop or something like that?

akdetrick: I missed that, good call!

akdetrick (approved):
Love it - thanks for adding this!

Issue #1172, opened 2024-04-03
[Slider]


akdetrick:
### Short Description
TK

### Responsibilities

**This component should be responsible for the following:**
-
-
-

**This component should NOT be responsible for...**
-
-
-

### Props Signature

```
<Component
  propA={}
  propB={}
```

```
Component.propTypes = {
}
```

akdetrick:
Merged by @agreatscott

Closing.

PR #1214, opened 2024-04-29
feat(Slider): Add slider component

agreatscott:
for https://github.com/narmi/design_system/issues/1172

This adds the `<Slider>` input component. It is a dual thumb range slider.

I referenced the example in https://react-spectrum.adobe.com/react-aria/useSlider.html#useslider-hook

below is the Figma for the desired look of the Slider

image

and how it behaves as implemented: 

https://github.com/narmi/design_system/assets/4285085/b8ae636b-075f-4c8c-9836-dc386b60815b

### DIFF
--- a/src/Slider/Thumbs.tsx
+++ b/src/Slider/Thumbs.tsx
@@ -1,0 +1,69 @@
+import { useRef } from "react";
+import { VisuallyHidden, mergeProps, useFocusRing, useSliderThumb } from "react-aria";
+
+const getRailWidth = (lowerInputThumbProps, higherInputThumbProps) => parseFloat(higherInputThumbProps.style?.left.toString().slice(0, -1)) - parseFloat(lowerInputThumbProps.style?.left.toString().slice(0, -1));
+
+const useThumb = (index, trackRef, name, state) => {
+  const inputRef = useRef(null);
+  const { thumbProps, inputProps } = useSliderThumb({
+    index,
+    trackRef,
+    inputRef,
+    name
+  }, state);
+  const { focusProps } = useFocusRing();
+  return {
+    inputRef,
+    thumbProps,
+    inputProps,
+    focusProps,
+  }
+}
+
+const Thumb = ({ thumbProps, inputRef, inputProps, focusProps }) => {
+  return (
+    <div
+      {...thumbProps}
+      className="thumb"
+    >
+      <VisuallyHidden>
+        <input ref={inputRef} {...mergeProps(inputProps, focusProps)} />
+      </VisuallyHidden>
+    </div>
+  )
+};
+
+const Thumbs = (props) => {
+  const { state, trackRef, lowerName, higherName} = props;
+
+  const lowerThumb = useThumb(0, trackRef, lowerName, state);
+
+  const higherThumb = useThumb(1, trackRef, higherName, state);
+
+  const width = getRailWidth(lowerThumb.thumbProps, higherThumb.thumbProps);
+  const activeRailStyle = {
+    left: lowerThumb.thumbProps.style?.left,
+    width: `${width}%`,
+  };
+
+  return (
+    <>
+      <Thumb
+        thumbProps={lowerThumb.thumbProps}
+        inputRef={lowerThumb.inputRef}
+        inputProps={lowerThumb.inputProps}
+        focusProps={lowerThumb.focusProps}
+      />
+      <div className="rail-active" style={activeRailStyle}/>
+      <Thumb
+        thumbProps={higherThumb.thumbProps}
+        inputRef={higherThumb.inputRef}
+        inputProps={higherThumb.inputProps}
+        focusProps={higherThumb.focusProps}
+      />
+    </>
+
+  );
+}
+
+export default Thumbs;
--- a/src/Slider/Thumbs.tsx
+++ b/src/Slider/Thumbs.tsx
@@ -23,14 +23,14 @@ const useThumb = (index, trackRef, name, state) => {
   }
 }
 
-const Thumb = ({ thumbProps, inputRef, inputProps, focusProps, isFocusVisible }) => {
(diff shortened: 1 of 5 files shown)


src/Slider/Thumbs.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const useThumb = (index, trackRef, name, state) => {
+  const inputRef = useRef(null);
+  const { thumbProps, inputProps } = useSliderThumb({
+    index,
+    trackRef,
+    inputRef,
+    name
+  }, state);
+  const { focusProps } = useFocusRing();
```
akdetrick, line 14: This looks keyboard navigable (nice!) but we should visually indicate focus somehow.

It would be good to sync with @hillarypollak on this to see if there's a good way to present a focus/active indicator for the currently focused slider handle.

My proposal would be a transparent theme primary ring around the handle dot while it's active.

src/Slider/Thumbs.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+const Thumb = ({ thumbProps, inputRef, inputProps, focusProps }) => {
+  return (
+    <div
+      {...thumbProps}
+      className="thumb"
+    >
+      <VisuallyHidden>
+        <input ref={inputRef} {...mergeProps(inputProps, focusProps)} />
```
akdetrick, line 30: We should aria-label this input "minimum"/"maximum" - maybe a `thumbLabel` prop that can be passed in.

I usually don't worry too much about translations in aria attributes, but if we want to enable that later, we could expose "minLabel"/"maxLabel" as props on `Slider`.

akdetrick (changes requested):
This is looking good! Just a few things to take care of:
- Lint errors (for internal components like `Thumb` you can eslint ignore the warnings)
- Visual indication of focus/active slider handle (I missed this in design review)
- `aria-label` in Thumb inputs
- some minor style stuff

The accessibility features are pretty nice - I get a good read of input values as I use the arrow keys to move the slider thumbs, but it's unclear which one is being moved.

Looks good though!

src/Slider/index.scss
```diff
@@ -0,0 +1,39 @@
+.nds-slider {
+  position: relative;
+  width: 300px;
```
akdetrick, line 3: The rail width is bounded by this `300px` right? Or is this the initial width?
Would it be possible to make this width auto or `100%`?

src/Slider/Thumbs.tsx
agreatscott: agreed on this

src/Slider/index.scss
agreatscott: ah yes good catch

src/Slider/index.scss
```diff
@@ -0,0 +1,39 @@
...
+    width: 16px;
+    height: 16px;
+    border-radius: 50%;
+    box-sizing: border-box;
+    background-color: var(--theme-primary);
+  }
+
+  .output-container {
+    margin-bottom: var(--space-l);
+  }
```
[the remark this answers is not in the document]
agreatscott: so yeah this was another thing i should have brought up in the PR. 

The text displayed in this `<output>` is what's adjacent to the slider in the designs (designs below)

image

it's not a label, since it's describing the values of the input, not what the input is for. For that reason, I've also included a visually hidden `<label>`, which serves to provide an accessible label for the input. This might be confusing to a consumer of this NDS component though. 

But to answer this question - this is styled as the designs, at least for the only usage of this component that i've seen. So maybe keep it as is? 

open to thoughts about the output/label though.

src/Slider/Thumbs.tsx
agreatscott: added minimum/maximum hardcoded (for now, can maybe add a prop later)

agreatscott:
Thanks for the review, I think I've given a decent go at resolving these comments. Only thing is I'm not sure the focus/hover ring is visible enough but I'll take another look in the morning

src/Slider/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  label: string;
+  output?: string;
+  higherName?: string;
+  lowerName?: string;
+  formatOptions: Intl.NumberFormatOptions;
+  maxValue: number;
+  defaultValue?: number[];
+  step: number;
+  value?: number[];
+  setValue?: CallableFunction
```
akdetrick, line 16: Do we need a `onChange` prop here?

```
<Slider
   value={valueStateVar}
   onChange={({ min, max }) => setValueStateVar([min,max])}
   ✂️ 
```

src/Slider/index.scss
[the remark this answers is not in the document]
akdetrick: Makes sense to me! I think using `output` is a nice choice here

akdetrick:
🎉 This is really easy to use with voiceover!

src/Slider/index.tsx
agreatscott: yep you're right - changed this to `onChange`.

src/Slider/Thumbs.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const Thumb = ({ thumbProps, inputRef, inputProps, focusProps, isFocusVisible, ariaLabel }) => {
+  return (
+    <div
+      {...thumbProps}
+      className={cc(["thumb", { "focus": isFocusVisible }])}
+    >
+      <VisuallyHidden>
+        <input ref={inputRef} {...mergeProps(inputProps, focusProps)} aria-label={ariaLabel} />
```
akdetrick, line 33: I think this is where we want the `onChange` to come from - `Thumbs` can fire `onChange` with the value of both inputs, and then `Slider` will be able to surface that to consumers

akdetrick (approved):

Issue #1174, opened 2024-04-03
[TokenInput] Delimiter-driven chip/token selection (freeform text)


akdetrick:
### Short Description

- Like gmail "to"
- Typeable text area that creates chips based on a delmiter
- Preferably editing caret goes at the end of the chips list, but could also be at beginning or newline
- default autogrow input to 2 rows
- backspace deletes chips (if applicable)

### Responsibilities

**This component should be responsible for the following:**
-
-
-

**This component should NOT be responsible for...**
-
-
-

### Props Signature

```
<Component
  propA={}
  propB={}
```

```
Component.propTypes = {
}
```

PR #1217, opened 2024-04-30
Feat/1174 token inputs

akdetrick:
relates to #1174 

- Adds `FieldToken`
- Adds `TokenInput`

### DIFF
--- a/src/TokenInput/index.jsx
+++ b/src/TokenInput/index.jsx
@@ -52,6 +52,11 @@ const TokenInput = ({
     const newTokens = new Set(tokens);
     newTokens.delete(token);
     onTokensChange([...newTokens]);
+
+    // always refocus the input after token list updates
+    if (inputRef?.current) {
+      inputRef.current.focus();
+    }
   };
 
   // The actual input element may wrap or shrink as tokens are added.
@@ -77,8 +82,6 @@ const TokenInput = ({
             className="nds-tokenInput-tokens margin--right--s"
             role="list"
             aria-label="Selected Items"
-            aria-live="polite"
-            aria-atomic="true"
           >
             {tokens.map((label) => (
               <FieldToken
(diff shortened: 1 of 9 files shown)


akdetrick:
I guess I should probably put a max width on tokens

Screenshot 2024-04-30 at 1 49 34 PM

akdetrick:
If the max width of `300px` for `FieldToken` becomes a problem, we can always change `startContent` in `Input` to have a max width with overflow hidden.

agreatscott:
I might be doing something wrong, but it looks like when I try to add new values to the input in the story, I can't get them to convert to tokens. But when I hardcode the value in they are of course shown as tokens, is it just that this is expected to be a controlled component? We almost always use controlled components for these inputs so I think this might be fine. 

image

agreatscott:
> `TokenInput` is a **fully controlled** component
 
 Ah, I just can't read! Disregard my above comment

src/index.scss
```diff
@@ -1,5 +1,10 @@
 @use "sass:map";
 
+// 💁 Imports are ordered by least specific to most specific.
+//    For example, helper classes should be further down in the cascade
+//    than base styles, so that selectors of the same specificity will
+//    resolve based on source order (last wins).
+
```
agreatscott, line 7: 🧠

src/TokenInput/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+import React, { useRef } from "react";
+import PropTypes from "prop-types";
+import FieldToken from "../FieldToken";
+import TextInput from "../TextInput";
+
+const noop = () => {};
+const MIN_TOKEN_LENGTH = 1; // assume anything less than 2 is an accident
+const INPUT_RESET_EVENT = { target: { value: "" } };
+
+// take value of input with any leading or trailing delimiters removed
```
akdetrick, line 15:
```suggestion
// take value of input with leading and trailing delimiters removed
```

agreatscott (approved):
discussed renaming some vars and accessibility tweaks offline - lgtm!

PR #1218, opened 2024-05-01
fix(IconButton): use default browser outline styling on focus-visible

elliot-wilson:
If you tab through a page on Chrome or Edge, IconButtons don't show evidence that they're being focused on. This is because they explicitly their set `outline` value to `inherit`. In practice, that means that they all have `outline: none`, even on `:focus-visible`.

I'm cool with us keeping the inherit behavior for the regular state, but unless there's a strong reason not to, on `:focus-visible`, let's revert the `outline` back to the default browser stylesheet.

### Before (these are focused, I promise)

image

image

### After

image

Screenshot 2024-05-01 at 10 30 38 AM

### DIFF
--- a/src/IconButton/index.scss
+++ b/src/IconButton/index.scss
@@ -17,6 +17,10 @@
       transform: translate(0px, -1px);
     }
   }
+
+  &:focus {
+    outline: revert;
+  }
 }
 
 .nds-icon-button--action {
--- a/src/IconButton/index.scss
+++ b/src/IconButton/index.scss
@@ -18,7 +18,7 @@
     }
   }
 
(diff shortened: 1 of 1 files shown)


akdetrick:
This would be a really good change for usability and accessibility.

We will be able to continue making custom focus rings for specific components, but this change ensures any interactive element without a custom focus ring uses the browser default style.

Code changes look good, but I'm going to sync with design to prepare them for the change and to see if there are any components that need a custom focus ring in the build of this PR.

akdetrick (approved):
I completely misread this initially. Thanks for the fix, missing focus indicators are the worst

PR #1227, opened 2024-05-06
feat(input): add a character counter to text inputs

kevmo314:
Adds these three states to `TextInput`:

image

They are located in the storybook under `TextInput > Examples`.

### DIFF
--- a/src/Input/index.js
+++ b/src/Input/index.js
@@ -82,7 +82,7 @@ const Input = ({
         {endContent && <div>{endContent}</div>}
       </div>
       <div className="nds-input-subline">
-        {/* this is in opposite direction to 1. make it easier to render and 2. accommodate screen reading order better */}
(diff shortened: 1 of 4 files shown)


src/Input/index.js
```diff
@@ -80,7 +81,13 @@ const Input = ({
         {(endIconClass || showClearButton) && endIconJsx}
         {endContent && <div>{endContent}</div>}
       </div>
-      <Error error={error} />
+      <div className="nds-input-subline">
+        {/* this is in opposite direction to 1. make it easier to render and 2. accommodate screen reading order better */}
```
elliot-wilson, line 85:
```suggestion
        {/* this is styled using row-reverse to 1. make it easier to render and 2. accommodate screen reading order better */}
```
Feel free to ignore! I wasn't sure what "in opposite direction" meant until I looked at the CSS

elliot-wilson (approved):

Issue #1237, opened 2024-05-07
`TextInput` - error prop is not passed down to native element


agreatscott:
### Component/Module

`TextInput`

### Reproduction Steps

add an error prop to `TextInput`
inspect native element, error prop is not defined on the native element

### Expected Result

error prop should be passed down to native element

### Actual Result

it's not

### Additional Details / Context

PR #1238, opened 2024-05-07
fix(TextInput):  prop was not being passed down to native input element

agreatscott:
resolve https://github.com/narmi/design_system/issues/1237

The moral of this is not to rely on prop-spreading - in https://github.com/narmi/design_system/commit/5e8de462cf642cbbe87e2d2489ffcf83f38824d6, we added a seemingly innocent change to sometimes display an error based on exceeding a given `maxLength` prop. But in so doing we removed the error field from `nativeElementProps`, which is spread onto the `<input>` element. So the `<input>` element doesn't receive the error prop. 

The `error` attribute is nonstandard on the native input element though, so we should maybe reconsider being dependent on it.

### DIFF
--- a/src/TextInput/index.js
+++ b/src/TextInput/index.js
@@ -56,15 +56,15 @@ const TextInput = React.forwardRef((props, forwardedRef) => {
     </div>
   ) : null;
 
+  const inputError = error ||
+    (maxLength && inputValue.length > maxLength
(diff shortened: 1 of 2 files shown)


akdetrick (approved):

akdetrick:
Thanks for the fix

PR #1240, opened 2024-05-09
chore(Slider): improve story

agreatscott:
A quick PR to improve the `<Slider>` storybook story. Autodocs wasn't really working well with typescript components that define prop interfaces instead of using `prop-types`, but I seem to have fixed that by using `react-docgen`. Inspired by [this comment](https://github.com/storybookjs/storybook/issues/21007#issuecomment-1430273342)

before:
image

after:
image

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -2,7 +2,7 @@ const path = require("path");
 
 module.exports = {
   typescript: {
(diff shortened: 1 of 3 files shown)


akdetrick (approved):
😮 
I think you just fixed something that I worked on for way too long last December.
It looks like I can start migrating from propTypes to TS interfaces after this!

PR #1243, opened 2024-05-13
Feat/nds 253 combobox tokens

akdetrick:
- Updates `DropdownTrigger` to accept start and end content (useful for managing tokens)
- Adds `MultiSelect` component
- Updates behavior of `FieldToken` component, so it works within a dropdown trigger

### `useMultipleSelection`

This hook from downshift is a state manager for more than one `selectedItem`, because `useSelect` only manages a single `selectedItem`.

I'm using state reducers in both hooks to control the behavior we want when users interact with the options list OR the chips in the dropdown trigger.


src/FieldToken/index.jsx
```diff
@@ -31,7 +35,8 @@ const FieldToken = ({ label, onDismiss = noop, testId }) => {
         role="button"
         aria-label={`Remove ${label}`}
         tabIndex={0}
-        onClick={() => {
+        onClick={(e) => {
+          e.stopPropagation();
```
akdetrick, line 39: We don't want this click to bubble up to any parent, as it may trigger another event. The dismissal click should be isolated to `FieldToken`

src/MultiSelect/MultiSelectItem.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const MultiSelectItem = ({ children }) => <>{children}</>;
```
akdetrick, line 4: This component gathers information about items so we can have an interface like this:

```jsx
<MultiSelect>
    <MultiSelect.Item>My item jsx</MultiSelect.Item>
</MultiSelect>
```

Instead of accepting a big data structure like this: `<MultiSelect options={{ ...bigJsObject }} />`.

I'm exposing this `Thing.Item` interface for two reasons:
- This would be consistent with other components, like `Select` and `Tabs`
- This gives consumers full control over what gets rendered in each item

src/MultiSelect/index.js
```diff
@@ -0,0 +1,276 @@
...
+                      "nds-select-item--highlighted":
+                        highlightedIndex === index,
+                      "rounded--top": index === 0,
+                      "rounded--bottom": index === items.length - 1,
+                      "nds-select-item--hasGutter": hasSelectedItem,
+                    },
+                  ])}
+                  {...getItemProps({ item, index })}
+                  role="option"
+                  aria-selected={isSelected(selectedItems, item).toString()}
```
akdetrick, line 231: When using a `useMultiSelect` in combination with a `useSelect`, we get the incorrect `aria-selected` value on items in the listbox. Here I'm just overriding what the prop getter returns, using my own helper to determine if the item is selected or not.

The `role` prop is only needed here to expose the linter to it (it's otherwise added by `getItemProps`)

agreatscott (approved):
tested this locally and it works and is keyboard accessible 🥳

PR #1244, opened 2024-05-14
fix(MultiSelect): add selectedItems prop for fully controlled component

akdetrick:
- Set a default behavior for `fieldValue`
- Add `selectedItems` prop that may be used to make a fully controlled `MultiSelect`
- Update docs

### DIFF
--- a/src/MultiSelect/index.js
+++ b/src/MultiSelect/index.js
@@ -24,6 +24,12 @@ const isSelected = (selectedItems, item) =>
   selectedItems.map(itemToString).includes(itemToString(item));
 
 /**
+ * Gets full item element by its `value` prop
+ */
+const getSelectedItems = (values, items) =>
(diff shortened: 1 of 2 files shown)


src/MultiSelect/index.js
```diff
@@ -176,7 +183,12 @@ const MultiSelect = ({
 
   return (
     <div className="nds-multiselect" data-testid={testId}>
-      <input type="hidden" name={name} id={name} value={fieldValue} />
+      <input
+        type="hidden"
+        name={name}
+        id={name}
+        value={fieldValue || selectedItems.map(itemToString).join(",")}
```
akdetrick, line 190: `fieldValue` is now an optional prop - it's only responsible for setting the hidden input value, and now it has the default behavior of holding a comma separated list of values

rogerioshieh (approved):

PR #1249, opened 2024-05-20
docs(Tooltip): use simpler endContent example for TextInput instead of absolute positioned Tooltip

elliot-wilson:
Updates the documentation for `Tooltip` so that the `Tooltip` + `TextInput` example uses the `endContent` prop on `TextInput` rather than a more homespun absolute positioned element.

image


akdetrick (approved):
bless you @elliot-wilson

PR #1254, opened 2024-05-29
fix(TextInput): use value in initial state instead of passing through…

akdetrick:
… as spread props

closes NDS-305

### What is the bug?
`TextInput` takes spread props and passes them down to `Input`. 
The `TextInput` component is the one managing the state variable `inputValue`.

The problem is that `value` is not used to initialize state, and is drilled down to `Input` without `TextInput` actually using it, making the value on initial render opaque to the character counter.

### How was this bug surfaced?
This hasn't caused any harm until now. We added a character counter, and we [noticed some weird behavior](https://linear.app/narmi/issue/NDS-305/nds-textinput-maxlength-not-updating-on-page-refresh) on initial render, where the counter wasn't reflecting the initial value.

### What is the fix?
**Without changing the current behavior of TextInput ,** make the counter check if there is a `value` prop being passed through.

### Long term fix
Combine `TextInput` and `Input` into a single, fully controlled component.
See also: [Roadmap](https://github.com/narmi/design_system/wiki/Roadmap#-fixes)

### DIFF
--- a/src/Combobox/index.test.js
+++ b/src/Combobox/index.test.js
@@ -38,7 +38,7 @@ describe("Combobox", () => {
     it("returns all children when list is not filtered", () => {
       const actual = getVisibleChildrenByCategory(
         MOCK_DISPLAYED_ITEMS,
-        MOCK_CATEGORY_CHILDREN
+        MOCK_CATEGORY_CHILDREN,
       );
       expect(actual).toStrictEqual(MOCK_CATEGORY_CHILDREN);
     });
@@ -46,7 +46,7 @@ describe("Combobox", () => {
     it("returns only visible children when first child is filtered out", () => {
       const actual = getVisibleChildrenByCategory(
         [...MOCK_DISPLAYED_ITEMS].slice(1, 4), // items: dos, tres, quatro
-        MOCK_CATEGORY_CHILDREN // contains: uno, dos
+        MOCK_CATEGORY_CHILDREN, // contains: uno, dos
       );
       const expected = [MOCK_CATEGORY_CHILDREN[1]];
       expect(actual).toStrictEqual(expected);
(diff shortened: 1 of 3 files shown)


src/Combobox/index.test.js
```diff
@@ -184,16 +184,16 @@ describe("Combobox", () => {
...
       <Combobox label={LABEL} inputValue={value}>
         {STATE_ITEMS}
-      </Combobox>
+      </Combobox>,
     );
     const input = screen.getByPlaceholderText(LABEL);
     expect(screen.getByDisplayValue(value)).toBeInTheDocument();
     fireEvent.change(input, { target: { value: "New value" } });
-    expect(screen.getByDisplayValue(value)).toBeInTheDocument();
+    expect(screen.getByDisplayValue(value)).not.toBeInTheDocument();
```
akdetrick, line 197: This is a little embarrassing. 😳 
I wrote this combobox test to specifically check for the weird behavior this PR fixes, but my assertion was the opposite of what it should have been.

src/Combobox/index.test.js
```diff
@@ -183,17 +183,4 @@ describe("Combobox", () => {
...
-    render(
-      <Combobox label={LABEL} inputValue={value}>
-        {STATE_ITEMS}
-      </Combobox>
-    );
-    const input = screen.getByPlaceholderText(LABEL);
-    expect(screen.getByDisplayValue(value)).toBeInTheDocument();
-    fireEvent.change(input, { target: { value: "New value" } });
-    expect(screen.getByDisplayValue(value)).toBeInTheDocument();
-  });
```
akdetrick, line 198: this test is no longer relevant

akdetrick, line 198: It looks like I have to clean up Combobox a bit now that `value` works as intended.

src/Combobox/index.test.js
```diff
@@ -38,15 +38,15 @@ describe("Combobox", () => {
     it("returns all children when list is not filtered", () => {
       const actual = getVisibleChildrenByCategory(
         MOCK_DISPLAYED_ITEMS,
-        MOCK_CATEGORY_CHILDREN
+        MOCK_CATEGORY_CHILDREN,
```
akdetrick, line 41: looks like I ran a formatter over this file - i'm fine with leaving this in

joanszhu (approved):

PR #1257, opened 2024-05-31
feat(Drawer): make padding size configurable via prop

akdetrick:
closes NDS-175

- Remove padding from `drawer` class
- Use helper class to fix the controls at `xxl` (their original padding)
- Use `paddingSize` prop value to apply a padding helper class to the drawer content area

Screenshot 2024-05-31 at 3 35 32 PM


rogerioshieh (approved):
Looks great!

PR #1262, opened 2024-06-13
Feature: fontsize respects user preferences

akdetrick:
closes NDS-103

## Background
Browsers offer an accessibility feature that allows users to change their base font size:
Screenshot 2024-06-13 at 11 59 57 AM

What this setting actually does is update the user agent stylesheet with a different `font-size` value for `body`.

Until recently, we had so much legacy UI in `banking`, it wasn't worth making this work until we cleaned up a bit. Now that we're in a better state with legacy styles, it's time to respect user font size preferences, starting with the NDS base styles.

## Changes

- Update our `style-dictionary` config so that any font or radius related tokens are generated as `rem` units. This is where the CSS vars come from.
- Add `rem` scss helper to NDS for any value that needs to scale with font size (e.g. nudging icons with transform)
- Replaced hard coded `px` values in NDS with variables wherever possible
- Used `rem()` helper to convert `px` units that are non-standard, or use for pixel pushing positioning

## What to look for
In addition to any helpful feedback you can provide about the actual code changes, use the storybook preview posted to this PR to look for any visual bugs. Change your browser font size preferences to see changes reflected in NDS components.

## How I tested this against `banking`
It's possible to [cut a beta release of NDS](https://github.com/narmi/design_system/blob/main/README.md#publishing-a-beta-version) for a potentially risky change, so we may install it in our `banking` consumers for testing. In this case, the beta version is `3.40.0-beta.1`.

✅ I've already tested this beta release thoroughly in AO, OLB, and the support app. Bug [tickets have been filed](https://linear.app/narmi/search?q=%5Bfont-size%5D&tab=issues) for anything that looks odd at larger font sizes in `banking`. I found **only minor issues** that **only appear when the font size preference is set to large or very large**.

### DIFF
--- a/src/FormSection/index.js
+++ b/src/FormSection/index.js
@@ -9,15 +9,17 @@ const FormSection = ({ title, children }) => (
   <div className="nds-formSection fontFamily--body">
     <Row alignItems="center" gapSize="xs">
       <Row.Item shrink>
-        <h3 className="fontColor--secondary fontSize--xs fontFamily--body">{title}</h3>
+        <h3 className="fontColor--secondary fontSize--xs fontFamily--body">
+          {title}
+        </h3>
       </Row.Item>
       <Row.Item>
-        <div style={{ width: "100%", height: "var(--space-m)" }}>
+        <div style={{ width: "100%", height: "var(--font-size-default)" }}>
           <svg
             xmlns="http://www.w3.org/2000/svg"
             version="1.1"
             xmlnsXlink="http://www.w3.org/1999/xlink"
-            height="var(--space-m)"
+            height="var(--font-size-default)"
             width="100%"
           >
             <line
@@ -40,6 +42,6 @@ FormSection.propTypes = {
   /** Title of form section */
   title: PropTypes.string.isRequired,
   children: PropTypes.node.isRequired,
-}
+};
 
 export default FormSection;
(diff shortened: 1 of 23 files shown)


src/FormSection/index.js
```diff
@@ -9,15 +9,17 @@ const FormSection = ({ title, children }) => (
...
     <Row alignItems="center" gapSize="xs">
       <Row.Item shrink>
-        <h3 className="fontColor--secondary fontSize--xs fontFamily--body">{title}</h3>
+        <h3 className="fontColor--secondary fontSize--xs fontFamily--body">
+          {title}
+        </h3>
       </Row.Item>
       <Row.Item>
-        <div style={{ width: "100%", height: "var(--space-m)" }}>
+        <div style={{ width: "100%", height: "var(--font-size-default)" }}>
```
akdetrick, line 17: `--font-size` vars scale; `--space` vars do not.

This is intentional, as text zoom should not increase gutters

src/Input/index.scss
```diff
@@ -98,12 +98,12 @@
     pointer-events: none;
     font-family: var(--font-family-body);
     font-weight: 400;
-    font-size: 12px;
+    font-size: var(--font-size-s);
     color: var(--color-mediumGrey);
-    line-height: 1.25;
+    line-height: var(--font-lineHeight-bodyText);
```
akdetrick, line 103: This is equivalent - I'm just doing some housekeeping here 🧹 

Line heights in NDS are always unitless and relative, so these values will continue to scale properly with font size adjustments.

TMiller00 (approved):

PR #1274, opened 2024-06-25
add hasError prop to checkbox

trentslutzky:
https://linear.app/narmi/issue/AO-876/beneficiaries-qa

Beneficiaries designs call for setting an error state on a group of checkboxes, but checkbox only allows for an error string to enable the error state on them.

This adds `hasError` prop to the checkbox that sets the error className and causes the red outline if hasError is true.

image


akdetrick (approved):
nice!

PR #1279, opened 2024-06-27
feat(RadioButton): Add inline row kind

cs-paz:
Adds a row `kind` to the `RadioButtons` component.

Implementation Details:
Flex parent div with:

- `justify-content: space-between`

Usage:
`<RadioButtons kind="row" />`


akdetrick (approved):
Looks great, thanks! 🎉

PR #1292, opened 2024-07-23
fix(MenuButton): prevent react-aria scrolljacking

akdetrick:
closes NDS-390

- Use `react-laag` `useLayer` hook for popover instead of the `react-aria` Popover. This avoids problematic behaviors from `react-aria`, like scrolling top and locking body scroll.
- Refactor use of `closePopover` so that the `onAction` of the Menu component is called before the popover closes

Tested against `banking` in `3.46.0-beta.{patch}` releases.


graphicsforge (approved):
Oh, it was react-aria's Popover that was the culprit!!!

PR #1302, opened 2024-07-29
fix(Drawer): add classname to navigation icons in mobile drawer

blooies:
resolves BANKW-821

- `X`, `chevron-left` and `chevron-right` icons under `drawer--vertical--mobile` were being affected by styles meant specifically for the navigation icon buttons
- added a classname `mobile-navigation-button` to the navigation icon's parent in order to make the styles more targeted 
- this is also consistent with how the navigation icons exist in non-mobile view where the parent has a `navigation-button` classname

### DIFF
--- a/src/Drawer/index.js
+++ b/src/Drawer/index.js
@@ -176,15 +176,15 @@ const Drawer = ({
         <>
           {showControls && (
             <>
-              <div onClick={onPrev}>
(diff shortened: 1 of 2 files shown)


akdetrick (approved):
👍 scopes CSS targeting an NDS icon

PR #1306, opened 2024-08-06
feat(icons): add up-down arrow icon

elliot-wilson:
image

Adds up/down arrow by rotating the transfers icon 90 degrees. We need this for indicating sortability on table columns in the staff portal.

Also updates the README, since I had a little trouble understanding where to download the icomoon zip.


src/icons/style.css
```diff
@@ -1,15 +1,18 @@
 @font-face {
-  font-family: "icomoon";
-  src: url("fonts/icomoon.woff?edoxd9") format("woff");
+  font-family: 'icomoon';
+  src:
+    url('fonts/icomoon.woff2?18z8tb') format('woff2'),
+    url('fonts/icomoon.ttf?18z8tb') format('truetype'),
+    url('fonts/icomoon.woff?18z8tb') format('woff'),
+    url('fonts/icomoon.svg?18z8tb#icomoon') format('svg');
```
akdetrick, line 7: Thanks to a bug with nextjs static file serving, we inline the font as base64 when building the NDS dist. To keep the size of this file reasonable, we serve only the `woff`.

```suggestion
  src: url('fonts/icomoon.woff?18z8tb') format('woff');
```

akdetrick (changes requested):

src/icons/style.css
elliot-wilson: just updated the CSS!

akdetrick (approved):

PR #1307, opened 2024-08-08
feat(TextInput): allow component to control native HTML 'required' attribute

elliot-wilson:
Previously, the `input` underlying the `TextInput` component was always set to `required`.

This led to unexpected browser behavior such as this pop-up for a field that wasn't actually required:

image

This PR preserves the existing behavior -- defaulting the input to be required -- while allowing individual `TextInput` instances to override it as needed.

### DIFF
--- a/src/TextInput/index.js
+++ b/src/TextInput/index.js
@@ -88,7 +88,7 @@ const TextInput = React.forwardRef((props, forwardedRef) => {
             value={inputValue}
             onChange={_onChange}
             onBlur={_onBlur}
-            required
+            required={props.required}
             placeholder={props.label}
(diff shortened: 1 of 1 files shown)


src/TextInput/index.js
```diff
@@ -104,7 +104,7 @@ const TextInput = React.forwardRef((props, forwardedRef) => {
           onBlur={_onBlur}
           ref={forwardedRef}
           type={type}
-          required
+          required={props.required}
```
akdetrick, line 107: technically we can already override this via spread props, but I like the more explicit prop here

akdetrick (approved):

Issue #1309, opened 2024-08-09
`Select` error text is too small


agreatscott:
### Component/Module

`Select`

### Screenshots (if applicable)

![image](https://github.com/user-attachments/assets/5422251f-f923-42d6-b566-b60838686e3e)

### Expected Result

`Select` error text should be consistent with other Input errors

### Actual Result

`Select` error text is too small

### Additional Details / Context

This is only really noticeable when seeing the component next to another NDS input component

PR #1310, opened 2024-08-09
fix(Select): fix text size for Select error

agreatscott:
close https://github.com/narmi/design_system/issues/1309

Hard to tell from this screenshot, but the error text for the `Select` component is now consistent with other input components (namely, `TextInput`)

I've also noticed that `TextInput` has more space between the input field and the error message - this is caused by the recent addition of the character counter. I think we're better off without that space, so I've removed it

image

<br />

image

### DIFF
--- a/src/DropdownTrigger/index.js
+++ b/src/DropdownTrigger/index.js
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import cc from "classcat";
+import Error from "../Error";
 
 /**
  * Generic trigger button for dropdowns. `DropdownTrigger` can be composed with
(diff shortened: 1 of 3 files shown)


src/DropdownTrigger/index.js
```diff
@@ -71,14 +72,7 @@ const DropdownTrigger = React.forwardRef(
...
         </div>
-        {errorText && (
-          <div className="nds-dropdownTrigger-error fontColor--error">
-            <span role="img" className="narmi-icon-x-circle fontSize--s" />
-            <span className="padding--left--xxs fontColor--error fontSize--xs">
-              {errorText}
-            </span>
-          </div>
-        )}
+        <Error error={errorText} />
```
akdetrick, line 75: ❤️

akdetrick (approved):

PR #1314, opened 2024-08-21
fix(Combobox, Select): color both input and dropdown borders on error

elliot-wilson:
I noticed that if you supply an `errorText` prop to a Combobox with children, the error highlighting doesn't apply to the border of the input.

image

This is different from the behavior for Select, for example:

image

As of this PR, the red border will apply to both the input and dropdown borders for both Combobox and Select:

image

image

image

I somewhat blindly followed some existing patterns, e.g.

https://github.com/narmi/design_system/blob/e75a10a84b6fd9f9d4296d0450f749d206f78f53/src/Select/index.js#L277-L281

and

https://github.com/narmi/design_system/blob/a468bba33e847a01f07b79197ef2f3b82bf42ce3/src/Combobox/index.js#L256-L262

I'm open to suggestions for other ways of doing this

### DIFF
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -404,7 +404,7 @@ const Combobox = ({
             {...getMenuProps(layerProps)}
             style={{
               width: triggerBounds?.width || "auto",
-              transform: `translateY(${layerSide === "top" ? "0px" : "-22px"})`,
+              transform: `${errorText ? `translateY(${layerSide === "top" ? "0px" : "-22px"} )` : "none"}`,
               ...layerProps.style,
             }}
           >
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -398,6 +398,7 @@ const Combobox = ({
                   isOpen && displayedItems.length > 0,
                 "nds-combobox-list--bottom": layerSide === "bottom",
                 "nds-combobox-list--top": layerSide === "top",
+                "nds-combobox-list--error": !!errorText,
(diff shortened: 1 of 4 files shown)


src/Combobox/index.js
```diff
@@ -403,6 +404,7 @@ const Combobox = ({
             {...getMenuProps(layerProps)}
             style={{
               width: triggerBounds?.width || "auto",
+              transform: `${errorText ? `translateY(${layerSide === "top" ? "0px" : "-22px"} )` : "none"}`,
```
elliot-wilson, line 407: I realize that this has a smell. I might be overlooking a more conventional way of doing this!

elliot-wilson:
> ideally we want the whole border here to be the error red vs mixing primary and red.

Yeah, I wondered about that! I've updated the PR so that it applies to the border to the dropdown, too -- for both Combobox and Select.

src/Combobox/index.js
akdetrick: This expression isn't very complicated. If we find ourselves doing more work than this in the future, we could do something like:

`getMenuStyle(triggerBounds, errorText, layerSide) // returns {}`

```
style={{
    ...getMenuStyle(triggerBounds, errorText, layerSize),
    ...layerProps.style
}}
```

✅ For now, I don't think any of the logic here is complicated enough to justify adding the indirection

akdetrick (approved):

PR #1315, opened 2024-08-22
feat(Checkbox): add condensed Checkbox

elliot-wilson:
Top margin was added to normal checkboxes that are subsequent siblings to normal checkboxes in https://github.com/narmi/design_system/pull/1089/files#diff-cfd872c173d6e6389363869af77b48becfc03fed480f697ec12eee4faea06f6aR7-R10

This results in spacing that is too large for our "filter chip" dropdowns, so we have to use custom styling to fend off the extra margin.

What we want:
image

What we get with the default:
image

To preserve backwards compatibility, I've added a new "condensed" kind that is like "normal", but without the extra top margin when there are multiple checkboxes. In my heart of hearts, I think the "normal" margin should be the one that changes, but I don't want to break anything.

Here's the Storybook in action:

https://github.com/user-attachments/assets/82909764-6ac9-4c18-be2f-fe3fb5158745

### DIFF
--- a/src/Checkbox/index.js
+++ b/src/Checkbox/index.js
@@ -164,7 +164,7 @@ Checkbox.propTypes = {
   /**
    * `normal` - visually matches a checkbox input
    *
-   * `condensed` - like `normal`, but without added top margin
+   * `condensed` - like `normal`, but without added top margin when there are multiple checkboxes
    * 
    * `card` - visually present as a toggleable card
    */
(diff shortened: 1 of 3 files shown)


akdetrick (approved):
Nice! This is 100% the preferred way to handle the design variation

hillarypollak:
this one looks good

PR #1316, opened 2024-08-22
feat(Combobox): allow custom end content

elliot-wilson:
Design would like to have a Combobox that doesn't have a up/down chevron indicating whether the Combobox is open/closed.

To address that in a more extensible way, I propose letting the consumer supply a `renderEndContent` function that receives an `isOpen` prop.

image

### DIFF
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -121,6 +121,7 @@ const Combobox = ({
   errorText,
   icon,
   testId,
+  inputEndContent,
 }) => {
   const allChildren = React.Children.toArray(children);
   const hasCategories = allChildren.some(
@@ -363,11 +364,12 @@ const Combobox = ({
           value={inputValue}
           startIcon={icon}
           endContent={
+            inputEndContent === undefined ? (
             <span
               className={`fontSize--l fontColor--primary narmi-icon-${
                 isOpen ? "chevron-up" : "chevron-down"
               }`}
-            />
+            />) : inputEndContent
           }
           {...getInputProps({
             onFocus: () => {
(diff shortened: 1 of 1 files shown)


src/Combobox/index.js
```diff
@@ -363,11 +364,12 @@ const Combobox = ({
           value={inputValue}
           startIcon={icon}
           endContent={
+            inputEndContent === undefined ? (
```
akdetrick, line 367: I was going to suggest a prop default, but I see the scope of the render is needed to decide the icon

akdetrick (commented):
@elliot-wilson Do you think we'll ever need to make the end content aware of the open state?

If so, we should do this as a function instead:

`endContent={(isOpen) => <h1>this is {isOpen ? "open" : "closed"}, lol</h1>}`


Love the approach here either way

elliot-wilson:
> @elliot-wilson Do you think we'll ever need to make the end content aware of the open state?
> 
> If so, we should do this as a function instead:
> 
> `endContent={(isOpen) => <h1>this is {isOpen ? "open" : "closed"}, lol</h1>}`
> 
> Love the approach here either way

oh, yes, you're exactly right -- a function is the way to go. I'll update this!

src/Combobox/index.js
```diff
@@ -121,6 +135,7 @@ const Combobox = ({
   errorText,
   icon,
   testId,
+  renderEndContent = defaultRenderEndContent,
```
akdetrick, line 138: **Optional**

Renaming this `endContent` would make it easier to remember (although types will help in the future anyway)

akdetrick: oh, that's why:

```
* Signature: `(isOpen) => React.ReactNode`
```

akdetrick (approved):

Issue #1318, opened 2024-08-22
Typescript types are not generated in a usable manner


agreatscott:
The types generated by NDS typescript are not in a format readily useable by consumers. We should remedy that.

PR #1319, opened 2024-08-22
chore(types): adjust tsconfig and include index.js in types

agreatscott:
for https://github.com/narmi/design_system/issues/1318

~The changes to the component files are just to add `any`, which the new ts config was complaining about (we should actually define these types at a later point)~ I was able to avoid these changes by not setting `strict: true` in the tsconfig. We probably want to enable strict mode in the future. 

I published this code as a beta version on NPM `3.48.0-beta.1`, and confirmed that I'm able to view types in VS Code.

Notable changes to get this to work:
- Replacing the `outFile` in the tsconfig to `"outDir": "./dist/types",`
- creating an `index.ts` file at the root source, and including all typescript components there. I initially tried to just convert the existing `index.js` to typescript, but ran into trouble when the generated typings file tried to look for types in all the components, and I couldn't figure out an easy way to ignore or create an empty declaration file for the components still implemented in JavaScript. 
- Made sure to only `include` the typescript files in the tsconfig

As we convert the components to typescript, we'll have to add them to the `index.ts` and the `include` configuration, which is another step but not too bad. When they're all converted, we can remove the `index.js`, set the root source to `index.ts`, and simply include all typescript files.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,7 +12,6 @@
     ],
     "moduleResolution": "node10",
     "sourceMap": true,
-    "strict": true,
     "target": "ES6",
     "skipLibCheck": true,
     "esModuleInterop": true,
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,7 +10,8 @@
         "ES2022",
         "dom"
     ],
-    "moduleResolution": "node10",
+    "moduleResolution": "NodeNext",
+    "module": "NodeNext",
     "sourceMap": true,
     "target": "ES6",
     "skipLibCheck": true,
@@ -18,25 +19,8 @@
     "jsx": "react-jsx",
     "emitDeclarationOnly": true,
     "outDir": "./dist/types",
-    "ignoreDeprecations": "5.0"
   },
   "files": [
-    "./src/index.ts",
-    "./src/types/Icon.types.ts",
-    "./src/Slider/Thumbs.tsx",
-    "./src/Slider/index.tsx",
-    "./src/SeparatorList/index.tsx"
-  ],
-  "include": [
-    "src/Alert/index.d.ts",
-    "src/index.ts",
-    "src/types/*.ts",
-    "src/Slider/*.tsx",
-    "src/SeparatorList/*.tsx"
-  ],
-  "exclude": [
-    "node_modules",
-    "dist",
-    "**/*.stories.tsx"
+    "src/index.ts"
   ]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,16 +1,43 @@
 {
   "compilerOptions": {
-    "outFile": "./dist/types/index",
-    "target": "ES6",
-    "lib": ["ES2022", "dom"],
-    "allowJs": true,
-    "forceConsistentCasingInFileNames": true,
-    "module": "NodeNext",
-    "moduleResolution": "NodeNext",
-    "jsx": "react-jsx",
+    "alwaysStrict": false,
     "declaration": true,
-    "emitDeclarationOnly": true
+    "declarationMap": true,
+    "importHelpers": true,
+    "inlineSources": true,
+    "isolatedModules": true,
+    "lib": [
+        "ES2022",
+        "dom"
+    ],
+    "moduleResolution": "node10",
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -1,16 +1,42 @@
...
+    "declaration": true,
+    "declarationMap": true,
+    "importHelpers": true,
+    "inlineSources": true,
+    "isolatedModules": true,
+    "lib": [
+        "ES2022",
+        "dom"
+    ],
+    "moduleResolution": "node10",
```
akdetrick, line 13: **Question**

Was setting a specific node runtime the fix here? (as opposed to an ES version/year?)

I struggled with this - there seems to be a few combinations between this config and the consumer that might work, but it seemed almost arbitrary.

tsconfig.json
```diff
@@ -1,16 +1,42 @@
...
-    "forceConsistentCasingInFileNames": true,
-    "module": "NodeNext",
-    "moduleResolution": "NodeNext",
+    "skipLibCheck": true,
+    "esModuleInterop": true,
     "jsx": "react-jsx",
-    "declaration": true,
-    "emitDeclarationOnly": true
+    "emitDeclarationOnly": true,
+    "outDir": "./dist/types",
```
akdetrick, line 20: 🙌 this is basically what I was trying to do in the first place!

tsconfig.json
```diff
@@ -1,16 +1,42 @@
...
   },
-  "include": ["**/*.ts", "**/*.tsx"],
-  "exclude": ["node_modules", "dist"]
-}
+  "files": [
+    "./src/index.ts",
+    "./src/types/Icon.types.ts",
+    "./src/Slider/Thumbs.tsx",
+    "./src/Slider/index.tsx",
+    "./src/SeparatorList/index.tsx"
```
akdetrick, line 28: **Question**

Why not just import everything we need in `./src/index.ts`?

tsconfig.json
```diff
@@ -1,16 +1,42 @@
...
+  ],
+  "include": [
+    "src/Alert/index.d.ts",
```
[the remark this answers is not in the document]
akdetrick: > As we convert the components to typescript, we'll have to add them to the index.ts and the include configuration, which is another step but not too bad.

I see 👍

akdetrick:
Is it okay that icon types are double nested? As long as consumers can pick them up when passing icon name props, it doesn't matter to me

```
└── types
    ├── SeparatorList
    │   ├── index.d.ts
    │   └── index.d.ts.map
    ├── Slider
    │   ├── Thumbs.d.ts
    │   ├── Thumbs.d.ts.map
    │   ├── index.d.ts
    │   └── index.d.ts.map
    ├── index.d.ts
    ├── index.d.ts.map
    └── types
        ├── Icon.types.d.ts
        └── Icon.types.d.ts.map
        ```

agreatscott:
> Is it okay that icon types are double nested?

Yeah I think this is fine - the typing structure just follows that of code itself, and we happen to place the icon types inside their own `type` dir... we then import and export it in `index.ts` so (I checked) consumers can `import { IconType } from @narmi/design_system` as expected. We could probably flatten this if we wanted to but I suggest we do that in a future change.

We don't actually reference the `IconType` anywhere in NDS yet (or at least not in a type we export) but we should.

tsconfig.json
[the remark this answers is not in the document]
agreatscott: This was accidentally included and I've removed it.

tsconfig.json
agreatscott: yes this is a smarter approach and I've simplified how we include files in `tsconfig.json`

tsconfig.json
agreatscott: Did some experimenting and had trouble with `NodeNext`, but `Node16` worked. So that's what we'll go with for now.

akdetrick (approved):
Looks good to go!

The `src/index.ts` as a source file for types is a good solution while we migrate more components to typescript

akdetrick:
semantic-release did its job; I missed that this was a `chore` commit. I'll push a `refactor` commit and force a release.

PR #1348, opened 2024-09-16
feat(RadioButtons): add checkmark variant

akdetrick:
closes NDS-461

Adds `checkmark` variant of `RadioButtons`. Useful for popovers with selection.

Example of where this would be used:
Screenshot 2024-08-22 at 10 36 44 AM

Screenshot from storybook preview:
Screenshot 2024-09-16 at 7 00 38 PM

### DIFF
--- a/src/RadioButtons/index.js
+++ b/src/RadioButtons/index.js
@@ -100,7 +100,7 @@ const RadioButtons = ({
                 role="presentation"
                 className={cc([
                   "nds-radio",
-                  { "narmi-icon-check": kind === "card" },
+                  { "narmi-icon-check": ["card", "checkmark"].includes(kind) },
                 ])}
               />
             </div>
@@ -147,7 +147,7 @@ RadioButtons.propTypes = {
    *
    * `card` - display input and label as a toggleable card
    */
-  kind: PropTypes.oneOf(["normal", "row", "card"]),
+  kind: PropTypes.oneOf(["normal", "row", "card", "checkmark"]),
   /**
    * Error message. When passed, the `error` prop will
    * render the radio group in an error state.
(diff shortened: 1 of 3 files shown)


src/RadioButtons/index.scss
```diff
@@ -136,6 +136,48 @@
...
+    &.nds-radiobuttons-option--focused .nds-radio,
+    &.nds-radiobuttons-option--checked .nds-radio {
+      visibility: visible;
+    }
+    &:hover .nds-radio,
+    &.nds-radiobuttons-option--focused .nds-radio {
+      color: var(--color-lightGrey);
+    }
+    &.nds-radiobuttons-option--checked {
+      color: var(--theme-primary);
```
laurenharpole, line 171: Notice and appreciate the use of an FI primary color for the checkmark

src/RadioButtons/index.js
```diff
@@ -100,7 +100,7 @@ const RadioButtons = ({
                 role="presentation"
                 className={cc([
                   "nds-radio",
-                  { "narmi-icon-check": kind === "card" },
+                  { "narmi-icon-check": ["card", "checkmark"].includes(kind) },
```
laurenharpole, line 103: Preface: this is an absolute nit, pr looks awesome!

`includes` checks for string elements in an array which containing a certain string value. This works well when we're checking for `kind` in a unique array of strings, but what if in the future, we want to add another`narmi-icon-check` that includes the word "card" or "checkmark"?

laurenharpole, line 103: Had an initial nit, but after double checking some includes behaviors, retracted it! This looks awesome

laurenharpole (approved):
Approved, looks great to me

PR #1349, opened 2024-09-16
fix(Dialog): prevent overflow gradient text overlap

akdetrick:
Closes NDS-401

Automatically add padding to the bottom of the content area equal to the height of the gradient when content is overflowing. This prevents the gradient from covering up the last line of text when you scroll to the bottom of the content area.

### DIFF
--- a/src/Dialog/index.scss
+++ b/src/Dialog/index.scss
@@ -1,3 +1,7 @@
+.nds-dialog {
+  --overflow-gradient-height: 20px;
+}
+
 .nds-shim--dark {
   position: fixed;
   top: 0;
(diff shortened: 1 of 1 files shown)


src/Dialog/index.scss
```diff
@@ -1,3 +1,7 @@
+.nds-dialog {
+  --overflow-gradient-height: 20px;
```
akdetrick, line 2: CSS Custom Properties cascade, so here I'm setting a var that only works within the children of `nds-dialog`, which prevents polluting our nice list of global vars

elliot-wilson: 🧠

elliot-wilson (approved):
EZ

PR #1356, opened 2024-09-19
feat(PageHeader): add PageHeader component

akdetrick:
Closes NDS-520

Adds `PageHeader` component. See storybook preview commented below for design review and docs.

### DIFF
--- a/src/IconButton/index.js
+++ b/src/IconButton/index.js
@@ -21,8 +21,9 @@ const IconButton = ({
   testId = "nds-icon-button",
   label,
   name,
-  onClick = () => {},
-  type,
+  // eslint-disable-next-line no-unused-vars
+  onClick = (e) => {},
+  type = "button",
   as = "button",
   ...otherProps
 }) => {
--- a/src/PageHeader/index.scss
+++ b/src/PageHeader/index.scss
@@ -1,0 +1,53 @@
+.nds-pageHeader {
+  background-color: rgba(var(--theme-rgb-primary), var(--alpha-10));
+}
+
+%chevron {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  align-self: center;
+
+  .nds-pageHeader--hero & {
+    font-size: rem(24px);
+    width: rem(24px);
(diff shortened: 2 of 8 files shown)


src/IconButton/index.js
```diff
@@ -21,8 +21,9 @@ const IconButton = ({
   testId = "nds-icon-button",
   label,
   name,
-  onClick = () => {},
-  type,
+  // eslint-disable-next-line no-unused-vars
+  onClick = (e) => {},
```
akdetrick, line 25: stops the typescript complaints about number of args for this `onClick` prop

src/IconButton/index.js
```diff
@@ -21,8 +21,9 @@ const IconButton = ({
...
+  type = "button",
```
akdetrick, line 26: Set a sensible default here

akdetrick:
### Design notes

- Support truncation in breadcrumbs
   - Consider `PageHeader.Crumb` interface of children
- Check if we can include chevrons _inside_ row items for breadcrumbs, so they always come after the label


#### Example breadcrumbs
"Checking > Account Name could be really long and we would prefer truncation"

akdetrick:
### Proposed interface change

Per design feedback, we will need to individually control truncation behavior of breadcrumbs. This interface would also prevent the duck typing we're doing with the array-like interface

```
<PageHeader>
   <PageHeader.BackLink label="string" onClick={() => {}} />
   <PageHeader.Breadcrumbs>
      <PageHeader.Crumb label="Breadcrumb link" onClick={() => {}} />
      <PageHeader.Crumb label="Long ass account name that we need to truncate" truncate={true} />
      <PageHeader.Crumb label="Current page" />
   </PageHeader.Breadcrumbs>
   <PageHeader.Menu>
      <MenuButton>...</MenuButton>
   </PageHeader.Menu>
   <PageHeader.Actions>
      <Button kind="secondary" label="action 1" />
      <Button kind="primary" label="action 2" />
   </PageHeader.Actions>
</PageHeader>
```

PR #1365, opened 2024-09-23
feat(Error): enable displaying of multiple errors

graphicsforge:
resolves https://linear.app/narmi/issue/BANKW-1152/enable-display-of-error-arrays

Sometimes, more than a single error is present for a single field. When this happens, we may want to display all of the errors in a list:

Screenshot 2024-09-23 at 9 56 54 AM

### DIFF
--- a/src/Error/index.js
+++ b/src/Error/index.js
@@ -1,18 +1,37 @@
 import React from "react";
 import PropTypes from "prop-types";
 
-const Error = ({ error, marginTop = "xxs" }) => {
-  if (!error) return null;
+const ErrorLine = ({ errorLine, marginTop = "xxs" }) => {
+  if (!errorLine) return null;
   return (
     <div className={`nds-input-error margin--top--${marginTop}`}>
       <div className="fontSize--s margin--right--xxs narmi-icon-x-circle" />
-      {error}
+      {errorLine}
     </div>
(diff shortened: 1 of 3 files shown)


src/Error/index.js
```diff
@@ -1,18 +1,37 @@
...
+  marginTop: PropTypes.oneOf(["xxs", "xs", "s", "m", "l", "xl", "xxl", "none"])
+};
+
+const Error = ({ error, marginTop = "xxs" }) => {
+  if (!error) return null;
+  if (Array.isArray(error)) {
+    return (
+      <div className="nds-input-errorlist">
+        {error.map((errorLine, index) => (
+          <ErrorLine key={index} errorLine={errorLine} marginTop={index ? marginTop : "xxs"} />
```
xiexiewenting, line 24: nit: i feel like react yells at me if i use a mapped array index as the key "since it could change"

xiexiewenting (approved):

src/Error/index.js
graphicsforge: oh you're right! I absolutely should have used the errorLine as the key, lemme fix that

graphicsforge:
Screenshot 2024-09-23 at 10 15 38 AM

PR #1366, opened 2024-09-23
feat(timelineEvent): Add support for a tooltip to TimelineEvent

aundus:
ref FED-287 

Adds support for tooltip for the icon. Will allow for use in the fednow manager components.

### DIFF
--- a/src/TimelineEvent/index.js
+++ b/src/TimelineEvent/index.js
@@ -2,12 +2,13 @@ import React from "react";
 import PropTypes from "prop-types";
 import iconSelection from "src/icons/selection.json";
 import cc from "classcat";
+import ToolTip from "../Tooltip"
 
 export const VALID_ICON_NAMES = iconSelection.icons.map(
   (icon) => icon.properties.name
 );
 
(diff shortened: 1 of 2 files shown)


justinolianonarmi (approved):

src/TimelineEvent/index.js
```diff
@@ -79,6 +80,10 @@ TimelineEvent.propTypes = {
   /**
    * Timeline event content (any JSX)
    */
+  tooltip: PropTypes.string,
+  /**
+   * Hover tooltip content for the icon
+   */
```
akdetrick, line 86: jsdoc should be above proptype

```suggestion
  /**
   * Hover tooltip content for the icon
   */
  tooltip: PropTypes.string,
```

akdetrick (changes requested):
one small change; otherwise lgtm!

src/TimelineEvent/index.js
aundus: @akdetrick Updated!

akdetrick (approved):

PR #1370, opened 2024-09-23
fix(MenuButton): restore icon rendering

akdetrick:
closes NDS-577

The `MenuButtonItem` components passed into `MenuButton` get mapped to a react-aria MenuButton. When we refactored this component to use react-aria-components, we forgot to include rendering the `startIcon` prop where we render out the `menuItems`.

Now the `MenuButtonItem` only takes props and returns a fragment. The actual item rendering is in `MenuButton/index.jsx` now.

Screenshot 2024-09-23 at 1 13 24 PM


reshav-abraham (approved):

xiexiewenting (approved):
they're back !!

PR #1374, opened 2024-09-23
feat(Dialog): use condensed header in mobile viewports

akdetrick:
Closes NDS-570

Per design, makes the `Dialog` have a header style that matches our native mobile app when at small viewports.

Our `Dialog` has always taken over the full screen below the smallest viewport. Now the header changes visual display and we leave a little at the top of the screen so you can see the shim peek through

dialog-takeover


laurenharpole (approved):
Looks lovely, approved!

xiexiewenting (approved):
looks good!  🎉

PR #1385, opened 2024-10-01
fix(MultiSelect): prevent field autofocus

akdetrick:
closes NDS-378

We observed the `MultiSelect` component stealing focus on page load in the staff app.

It looks like the prop getters being used on the `DropdownTrigger` inside the multiselect were competing over binding refs and events.

This PR resolves the issue by:
- Creating a separate, dedicated element around the `DropdownTrigger` to receive the `ref` for the `userLayer` positioning engine
- Simplify the downshift props being passed to the `DropdownTrigger` by applying only the `getToggleButtonProps` from the `useSelect` hook

-----

### tl;dr

The `triggerProps` returned by the `useLayer` positioning engine does one thing, and only one thing: it applies a `ref` to an element so the element may be used as the positioning reference.

Wrapping the `DropdownTrigger` in a div does effectively the same thing, without being in competition with downshift prop getters attempting to apply a ref

### DIFF
--- a/src/MultiSelect/index.js
+++ b/src/MultiSelect/index.js
@@ -64,7 +64,6 @@ const MultiSelect = ({
   /** @see https://www.downshift-js.com/use-multiple-selection/#usage-with-select  */
   const {
     getSelectedItemProps,
-    getDropdownProps,
     addSelectedItem,
     removeSelectedItem,
     selectedItems,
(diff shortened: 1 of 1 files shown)


src/MultiSelect/index.js
```diff
@@ -191,22 +190,20 @@ const MultiSelect = ({
...
-      />
+      <div {...triggerProps}>
+        <DropdownTrigger
+          disabled={disabled}
+          isOpen={isOpen}
+          labelText={label}
+          startContent={renderTokens()}
+          errorText={errorText}
+          labelProps={{ ...getLabelProps() }}
+          {...getToggleButtonProps()}
```
akdetrick, line 201: This prop getter is returned by `useSelect`, which manages the state of the select popover.

The `useMultipleSelection` hook only needs to be responsible for managing the state of selected items, so I was able to remove the `getDropdownProps` getter. The `getDropdownProps` getter is only necessary if `useMultipleSelection` is being used in isolation (without `useSelect`).

wrbrand (approved):
Adam walked me through this and it now makes sense to me.

PR #1386, opened 2024-10-02
fix(TokenInput): stop clearing input on blur

akdetrick:
closes NDS-428

`TokenInput` originally cleared the input value on blur. This PR removes that behavior per design.


laurenharpole (approved):
Changes make a lot of sense and gave the story book page a look, approved!

Issue #1392, opened 2024-10-07
[Dialog] Header in desktop has extra padding on the left


xiexiewenting:
### Component/Module

Dialog

### Reproduction Steps

Open dialog in desktop size screen

### Expected Result
![Screenshot 2024-10-07 at 10 32 44 AM](https://github.com/user-attachments/assets/c3411a72-4a4f-4d36-9531-afb1148060cc)

### Actual Result
![Screenshot 2024-10-07 at 10 33 29 AM](https://github.com/user-attachments/assets/29dbe578-5c8e-4343-aa05-6ed9b9179210)

PR #1393, opened 2024-10-07
fix(Dialog): remove left padding from header in desktop mode

xiexiewenting:
resolves https://github.com/narmi/design_system/issues/1392


JohnathanWeisner (approved):
👨‍🍳

PR #1403, opened 2024-10-16
feat(Combobox): add Combobox.Action support

akdetrick:
Closes NDS-458

Adds support for passing a `Combobox.Action` to Combobox. Matches the same pattern we use in `Select.Action` components.

When a user clicks or selects an action via keyboard, the menu will close and the `onSelect` callback of the action item will be invoked.

When filtering by text, the actions will **always** be visible at the bottom of the list of suggestions.

Developers may place an Action item at any position of the list, including inside a category. When filtering, all actions will appear at the bottom of the list as usual.

See storybook preview posted below to test the new feature.

### DIFF
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -7,6 +7,7 @@ import { useLayer } from "react-laag";
 import ComboboxItem from "./ComboboxItem";
 import ComboboxHeading from "./ComboboxHeading";
 import ComboboxCategory from "./ComboboxCategory";
+import ComboboxAction from "./ComboboxAction";
 import TextInput from "../TextInput";
 import Row from "../Row";
 import { getItemIndex } from "../Select";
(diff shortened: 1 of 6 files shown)


src/Combobox/index.js
```diff
@@ -17,14 +18,26 @@ export const VALID_ICON_NAMES = iconSelection.icons.map(
...
+  }
+  return result;
+};
+
 /**
  * @param {Object} component a Combobox.Item or Combobox.Heading component
- * @returns {Boolean} true if the item is a selectable Combobox.Item
+ * @returns {Boolean} true if the item is a selectable Combobox.Item or Action
  */
 export const isSelectable = (component) => {
```
akdetrick, line 37: "selectable" means selectable via keyboard as part of the options list of the combobox

src/Combobox/index.js
```diff
@@ -182,11 +196,12 @@ const Combobox = ({
...
       // into `useCombobox` by filtering the initial items list from input value
       if (!disableFiltering) {
+        const actionItems = items.filter(isAction);
         const filteredItems = filterItemsByInput(
-          items.filter(isSelectable),
+          items.filter((item) => !isAction(item)).filter(isSelectable),
           inputValue.toLowerCase(),
         );
-        setDisplayedItems(filteredItems);
+        setDisplayedItems([...filteredItems, ...actionItems]);
```
akdetrick, line 204: This is how we ensure actions are always visible at the bottom when filtering

src/Combobox/index.js
```diff
@@ -17,14 +18,26 @@ export const VALID_ICON_NAMES = iconSelection.icons.map(
   (icon) => icon.properties.name,
 );
 
+/**
+ * @param {Object} item Combobox.{Action|Item|Heading} component
+ * @returns {Boolean} true if the item is a Combobox.Action
+ */
+export const isAction = (item) => {
```
graphicsforge, line 25: Oh wow, I'm behind the times: was initially thinking that this would be more literally an isAction function if we could enforce ComboboxAction.propTypes on item... but we can't do that anymore and they raise "Invariant Violation: Calling PropTypes validators directly is not supported by the `prop-types` package." and don't allow a way of calling PropTypes.checkPropTypes() in a way that returns or raises errors =(

I suppose we could call this getLabel() or hasLabel(), but so long as we don't add too many other Combobox elements it shouldn't pose too much of a problem

graphicsforge (approved):

src/Combobox/index.js
akdetrick: Yeah, I'm not crazy about this as it is. I'm making my way through components converting things to typescript, which should make some of these checks and validation much better

PR #1411, opened 2024-10-21
feat(CollapsibleCard): update disabled state; add renderTitle prop

akdetrick:
closes NDS-586

- Updates disabled state to use `DisabledShim` (per design)
- Adds `renderTitle` render function to allow complete customization within the clickable title area

We got a bit too specific with our content slots when shipping `CollapsibleCard`, and our design team has been experimenting with slightly different content for this component in some features. 

The `renderTitle` render prop allows developers to define whatever they want as the content of the title area. Implementing this as a render prop function allows the consumer to use the `isOpen` state of the card to conditionally render things.

Screenshot 2024-10-21 at 3 43 46 PM

### DIFF
--- a/src/CollapsibleCard/index.js
+++ b/src/CollapsibleCard/index.js
@@ -3,6 +3,9 @@ import PropTypes from "prop-types";
 import cc from "classcat";
 import Row from "../Row";
 import IconButton from "../IconButton";
+import DisabledShim from "../DisabledShim";
+
+const noop = () => {};
 
 const CollapsibleCard = ({
   title,
@@ -10,10 +13,11 @@ const CollapsibleCard = ({
   statusText,
   isOpen,
   trigger = "header",
-  onOpen = () => {},
-  onClose = () => {},
+  onOpen = noop,
+  onClose = noop,
   isDisabled = false,
-  onDisabledClick = () => {},
+  onDisabledClick = noop,
+  renderTitle,
   hasError = false,
   disableHover = false,
   children,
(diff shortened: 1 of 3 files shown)


src/CollapsibleCard/index.js
```diff
@@ -3,17 +3,21 @@ import PropTypes from "prop-types";
 import cc from "classcat";
 import Row from "../Row";
 import IconButton from "../IconButton";
+import DisabledShim from "../DisabledShim";
+
+const noop = () => {};
```
akdetrick, line 8: noop is a named function - sort of a micro optimization for JS garbage collection (`() => {}` functions are anonymous and redefined each pass)

src/CollapsibleCard/index.js
```diff
@@ -175,7 +185,6 @@ const CollapsibleCard = ({
           onTitleContainerClick(isDisabled, "open");
       }}
       aria-expanded="false"
-      aria-disabled={isDisabled ? "true" : "false"}
```
akdetrick, line 178: This attribute is now handled by `DisabledShim`

src/CollapsibleCard/index.stories.js
```diff
@@ -179,6 +179,28 @@ WithError.args = {
   ),
 };
 
+export const WithCustomTitle = Template.bind({});
+WithCustomTitle.args = {
+  renderTitle: (isOpen) => (
+    <div className="padding--all">
+      This card is <strong>{isOpen ? "open" : "closed"}</strong>.
+    </div>
+  ),
```
akdetrick, line 188: This is what gives us unlimited flexibility moving forward. Collapsible cards are used in a lot of our new designs, and we have a new designer joining soon too. Keep this component flexible will keep it alive and useful as we try out different stuff in design

blooies (approved):
lgtm!! thank you for adding this functionality to the card!

PR #1424, opened 2024-11-04
feat(CollapsibleCard): add radiusSize prop to configure border radius

akdetrick:
closes NDS-717

Allows border radius of `CollapsibleCard` to be configured via props. Sets "m" as default.

Screenshot 2024-11-04 at 12 10 19 PM


blooies (approved):
lgtm, thank you for adding!!

PR #1436, opened 2024-12-02
build: add timing to webpack build

### LINKED ISSUE #1413: [build] Introduce build timing
As part of our developer experience initiative, let's set up some build timing that we can use to measure the impact of our changes.

This should be against a full webpack build.

akdetrick:
closes #1413 

Uses linux `time` command to print timing of full webpack compile whenever the webpack CLI is called.

#### Example output

```
webpack 5.96.1 compiled with 3 warnings in 5647 ms
	6.23s real		11.89s user		0.69s sys
```


akdetrick:
Nope, not going to work. We'll need a webpack plugin. Will do this later.

PR #1437, opened 2024-12-02
fix(Tooltip): improve a11y on trigger element

akdetrick:
closes NDS-701

## Summary
Keyboard-only (and other assistive tech) users are able to navigate to the trigger of the tooltip element and operate it, which is good. However, the trigger element has a `role="button"`. There's no actual button action though; it only exists to create a focus/blur side effect for keyboard users.
This is also an issue because every tooltip trigger is getting added to navigation landmarks, which seems wrong.

[Loom video of this issue (🔉 sound on)](https://www.loom.com/share/482f0c14b9eb43e9a9dab2f31af1fbf0?sid=be5e906b-bd10-4c7a-88e4-c2aae4aa8b5c)

## Changes
All `Tooltip` trigger elements should be read as focusable text content by assistive tech.  The tooltip itself should be announced "politely" only when the tooltip appears.

- Remove `button` ARIA role from the popover trigger. It doesn't actually take an action and acts passively as it's focused, so it should __not__ have a button role.
- Adds `aria-expanded` to make it clear that the announced tooltip content is in a modal toolip

The WAI-ARIA working group[ doesn't endorse any specific tooltip pattern officially](https://github.com/w3c/aria-practices/issues/128), but this interaction pattern as implemented [matches the spec for a snackbar in Material Design](https://m3.material.io/components/snackbar/accessibility).


graphicsforge (approved):
ah, it's not a button but needs a tabindex

akdetrick:
> ah, it's not a button but needs a tabindex

Exactly - it's a weird solution, but a workable one for something that only happens on focus/blur rather than actual button press

PR #1438, opened 2024-12-03
fix(Icons): remove fill from feed, security icons

tlackemann:


akdetrick (approved):
Removing default fill allows us to recolor SVGs when using the SVG data directly (as we do in Atlas)

PR #1440, opened 2024-12-05
feat(Snackbar): add snackbar table status/toolbar component

akdetrick:
Closes NDS-521

Implements `Snackbar` (adapted from `FloatingPillBox` in the staff app) as a responsive component in NDS.

Consumers of this component are responsible for positioning the Snackbar, allowing greater flexibility between different use cases.

This differs from the staff app component in a few important ways:
- dividers change from vertical to horizontal based on viewport
- some flex items change their basis at mobile sizes, forcing them to a new line
- `aria-live` region and roles
- conditional rendering of the toolbar itself, which is used to announce the live region

### Accessibility

**aria-live attr**
Announces its content when its DOM subtree is modified.

**status value**
This live region value gives assistive tech a hint that this element contains a status, like "3 new emails". In this case, it's the status of user selection elsewhere on the page.
The announcement behavior is the same as "polite", which waits for a pause before announcing content.

**toolbar role**
It provides a special navigation landmark for an area with commonly used controls, making it easy to switch between the toolbar and other content.

### Layout

Component must conform to these two layouts in narrow and wide viewports.
The component should take any number of children and dividers, in any order.

Screenshot 2024-12-04 at 7 15 15 PM
Screenshot 2024-12-04 at 7 15 32 PM


src/Snackbar/index.scss
```diff
@@ -0,0 +1,41 @@
...
+  width: max-content;
+  min-width: 330px;
+  max-width: 80vw;
+  margin: 0 auto;
+
+  @include atMediaUp("m") {
+    flex-wrap: nowrap;
+    border-radius: rem(56px);
+    padding: var(--space-s) var(--space-m);
+  }
```
akdetrick, line 21: The "default" properties target _all_ viewport sizes, making mobile sizes essentially the default in our mobile-first media queries.

This `atMediaUp` generates a media query that targets "m" size viewports _and larger_.
Setting `nowrap` on the flex forces all children of Snackbar into one line.

src/Snackbar/index.scss
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.nds-snackbar-divider,
+.nds-snackbar-buttonGroup {
+  flex-basis: 100%;
+  @include atMediaUp("m") {
+    flex-basis: unset;
+  }
+}
```
akdetrick, line 30: `flex-basis` is used to force a flex item to wrap to the next line. The basis is relative to the parent, so `100%` does the trick.

This is how we make sure dividers and button groups get their own line while other flex children remain in row flow

src/Snackbar/index.scss
```diff
@@ -0,0 +1,41 @@
...
+
+.nds-snackbar-buttonGroup {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: var(--space-m);
+
+  .nds-button {
+    flex-shrink: 1;
+  }
```
akdetrick, line 40: each flex item should be no larger than its content width, making it easier to center buttons within their group

src/Snackbar/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Wraps text nodes with a dedicated element.
+ */
+const SnackbarText = ({ children }: { children: React.ReactNode[] }) => (
+  <div className="nds-snackbar-text fontColor--secondary">{children}</div>
+);
+
+/**
+ *
```
akdetrick, line 37:
```suggestion
 * Renders a semantic grouping of buttons within the Snackbar
```

src/Snackbar/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  </ul>
+);
+
+export interface SnackbarProps {
+  children: React.ReactNode[];
+  isActive: boolean;
+}
+
+/**
+ * A status region for multiple selection in a table.
```
akdetrick, line 53:
```suggestion
 * A status toolbar for multiple selection in a table.
```

src/Snackbar/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * A status toolbar for multiple selection in a table.
+ * Intended to be rendered in fixed position over a table.
+ */
+const Snackbar = ({ children, isActive = false }: SnackbarProps) => {
+  return (
+    <div aria-live="polite" role="status">
+      {isActive && (
```
laurenharpole, line 59: Everything I'm seeing here makes a lot of sense

blooies (approved):
lgtm! thank you for adding! love the added touch of how it turns horizontal on smaller screens!

kevmo314 (approved):
Neato!

laurenharpole (approved):
Reviewed the functionality today locally. So excited to see this in action, really like how responsive it is. LGTM!

PR #1446, opened 2024-12-10
Create a custom context menu

xian0831:
Implementation of ContextMenu.
- ContextMenu wraps other elements or components to enable a list of menu items. 
- The list of menu items can be open by right clicking on the wrapped element or press "Control + F12" when the wrapper element is focused.
- Since there is no consist way to open context menu from the OS level, instruction for opening context menu is included in aria-label.
- The implementation of this component is influence by MenuButton 

https://linear.app/narmi/issue/ADMIN-4019/[frontend]-add-right-click-menu


kevmo314 (approved):

akdetrick:
# Very nice!

PR #1449, opened 2024-12-13
Fix/nds 809 select action in category

akdetrick:
Closes NDS-809

I recently encountered a new requirement where Select components need both categories _and_ actions. The existing component has a bug where `Action` subcomponents end up getting grouped under a category without a title.

This PR changes the way we pull `children` from usage, so that `Select.Action` components are always rendered last, and outside any categories.


src/Select/index.js
```diff
@@ -58,9 +58,16 @@ export const getItemByValue = (value, items) => {
...
+export const getItemIndex = (item, items) => {
+  let result = 0;
+  if (isAction(item)) {
+    result = items
+      .map(({ props }) => props.onSelect)
+      .indexOf(item.props.onSelect);
+  } else {
+    result = items.map(({ props }) => props.value).indexOf(item.props.value);
+  }
+  return result;
```
akdetrick, line 70: These changes address an issue where `Select.Action` subcomponents weren't finding their index correctly.

src/Select/index.js
```diff
@@ -123,21 +130,27 @@ const Select = ({
...
+  actions = React.Children.toArray(children).filter(isAction);
 
-  // If categories are being used, extract items from categories
-  if (allChildren.some(({ type }) => type.displayName === "Select.Category")) {
-    items = allChildren.flatMap(({ props }) =>
-      React.Children.toArray(props.children),
-    );
-    categories = allChildren.map(({ props }) => ({
+  // If categories are being used, extract items and actions from categories
+  // FIXME: I fucked up here. I think this is the last bit I need to worry about.
```
akdetrick, line 142:
```suggestion
```

src/Select/index.js
```diff
@@ -123,21 +130,26 @@ const Select = ({
...
   let items = [];
   let categories = [];
+  let actions = [];
   const [userInput, setUserInput] = useState(""); // most recent val the user typed while focused on this input
 
-  const allChildren = React.Children.toArray(children);
+  const options = React.Children.toArray(children).filter(
+    (item) => !isAction(item),
+  );
+  actions = React.Children.toArray(children).filter(isAction);
```
akdetrick, line 139: Now we _always_ extract `actions` as separate from `options`. You can see below that when we assign the `items` var, it's a combination of `[...options,...actions]`. This array we set to the `items` var is essential to downshift state management.

Regardless of how and where we render these things, downshift needs to have a complete accounting of each item. Rendering matches the item order up with downshift in `renderItem`, which uses prop getters to apply the appropriate props to each item.

elliot-wilson (commented):
Makes sense to me!

PR #1456, opened 2024-12-16
fix(ContextMenu): fix missing export for ContextMenu in index.js

xian0831:
ContextMenu was included in `index.ts` but I didn't realize it also need to be export it in `index.js`. This fix should make sure ContextMenu can be imported properly.

Ticket: https://linear.app/narmi/issue/ADMIN-4066/[frontend]-fix-context-menu-export-issue


akdetrick (approved):
Sorry I missed this!

I may make a danger rule to help catch this because I've done it enough times myself

PR #1457, opened 2024-12-16
refactor(Row)!: convert Row and Row.Item to TypeScript

xian0831:
Staff portal uses `Row` pretty heavily. Moving `Row` and `Row.Item` can help IDE autocomplete props and prevent incorrect usage of those component.

https://linear.app/narmi/issue/NDS-815/[ts-convert]-convert-row-component-to-typescript


akdetrick (approved):
Looks good! Thanks for doing this.

I can confirm types are being built correctly and I do not see any regressions in component behavior

PR #1458, opened 2024-12-17
fix(disabledShim): Add full width to disabledShim

TMiller00:
For [AO-2050](https://linear.app/narmi/issue/AO-2050/staff-led-status-is-misaligned-header-color-incorrect). Add full-width to `disabledShim`


akdetrick (approved):
Fix makes sense - the disabledShim in the header was shrink wrapping contents, and the 100% width ensures it always fills its parent

PR #1462, opened 2024-12-20
fix(TypeScript): add a `json.d.ts` file to remove @ts-expect-error when importing JSON file

xian0831:
Previously, I suppress the TS error when importing a JSON file. Instead of doing that in all TS files, let's add a JSON module.    

Technically, we don't want to use `any` for JSON values, but I suspect we won't get too much value from making them precisely typed and the amount to work is non-trivial.    

https://linear.app/narmi/issue/NDS-837/[nds]-add-declaration-file-for-json


akdetrick (approved):
Incredible - love to see this issue finally resolved, thank you!

PR #1469, opened 2024-12-23
Breaking/nds 833 v4 deprecations

akdetrick:
closes NDS-833

We released major version 4 recently, but there were a few deprecation removals that didn't make it in.

This PR introduces some more breaking changes on v4 as a patch version, but I will update our changelog to say that this applies to anything v4 or above.

I will handle the NDS upgrade in `banking`. We only use these deprecated things in a few places, and we can always remap the removed CSS vars on the `banking` side.

### Changes
- remove deprecated CSS vars
- remove `Dropdown` component
- remove `defaultActive` prop from `Toggle`
- update minimum node version to `>= 22` (LTS)

### Follow up tasks

@akdetrick To update CHANGELOG manually after this merges, so these breaking changes are listed under the 4.0.0 release, with notes on how to handle the changes.


src/deprecations.json
```diff
@@ -1,16 +1,4 @@
...
-    "--nds-red",
-    "--nds-white",
-    "--nds-font-family",
-    "nds-header-font",
-    "subdued-20-opacity",
-    "subdued-10-opacity"
-  ],
-  "components": ["Dropdown"]
+  "cssCustomProperties": [],
+  "components": []
```
akdetrick, line 3: This file exists so that consumers can create their own danger rules and/or lint warnings when deprecated components or vars are used.

akdetrick:
Labelling `DO NOT MERGE` as there are important follow up tasks for me to do once this merges.

This should be author merged only.

akdetrick:
## Postponing this change

These changes should go in the next major version of NDS, planned roughly for February. Closing for now.

PR #1473, opened 2024-12-31
Fix/nds 834 dialog ios keyboard

akdetrick:
Closes NDS-834

Fixes ios Safari issue where the footer is cut off by the bottom of the viewport or keyboard.

This also brings `Dialog` within design specs for how height should behave on mobile sized screens - the Dialog should only be tall enough to accommodate its content. The height of the dialog should max out just below the top of the screen so that a little bit of shim is always visible.

### Before
Screenshot 2024-12-30 at 7 18 03 PM

### After
Screenshot 2024-12-30 at 7 19 25 PM
Dec-30-2024 19-21-44

## How / Why

The original bug was due to the `FocusLock` component adding a div between the shim and the dialog. 

The FocusLock div became the implicit flex child of the shim. We were setting constraints on the `nds-dialog` element when we should have been setting constraints on the shim, which creates a more reliable cross browser layout.

#### Shim layout responsibilities (`.nds-shim`)
- cover the entire viewport, within the mobile "safe area". The `env()` functions in CSS take a second argument as a fallback. We set them to inset `0px`.

#### FocusLock div layout responsibilities (`.nds-dialog-focuslock`)
- Absolutely positions itself to the full height of the "safe area" via `env()` functions for top and bottom position
- Set a flex column so that the Dialog is always justified to the bottom of the FocusLock div

#### Dialog div layout responsibilities (`.nds-dialog`)
- Allow the entire dialog to grow in height, but **only up to 96% of its parent container, the FocusLock div**

### DIFF
--- a/src/Dialog/index.js
+++ b/src/Dialog/index.js
@@ -96,7 +96,7 @@ const Dialog = ({
   const dialogJSX = (
     <CSSTransition timeout={1} classNames="nds-dialog-transition" appear in>
       <div className="nds-shim--dark" ref={shimRef} onClick={handleShimClick}>
-        <FocusLock autoFocus={false}>
+        <FocusLock autoFocus={false} className="nds-dialog-focuslock">
           <div
             role="dialog"
             aria-labelledby="aria-dialog-label"
--- a/src/Dialog/index.scss
+++ b/src/Dialog/index.scss
@@ -3,16 +3,13 @@
 }
 
 .nds-shim--dark {
-  position: fixed;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
+  position: absolute;
+  top: env(safe-area-inset-top, 0px);
+  right: env(safe-area-inset-right, 0px);
+  bottom: env(safe-area-inset-bottom, 0px);
+  left: env(safe-area-inset-left, 0px);
   background-color: var(--bgColor-scrimDark);
-  display: flex;
-  justify-content: center;
-  align-items: flex-end;
-  min-height: 100vh;
+  min-height: 100dvh;
   z-index: 3;
 
   // create spaces around viewport edges for tablet or larger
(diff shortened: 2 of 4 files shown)


src/Dialog/index.js
```diff
@@ -96,7 +96,7 @@ const Dialog = ({
   const dialogJSX = (
     <CSSTransition timeout={1} classNames="nds-dialog-transition" appear in>
       <div className="nds-shim--dark" ref={shimRef} onClick={handleShimClick}>
-        <FocusLock autoFocus={false}>
+        <FocusLock autoFocus={false} className="nds-dialog-focuslock">
```
akdetrick, line 99: Because this component inserts a div, we must use this element to set height constraints on the inner `nds-dialog` element

src/Dialog/index.scss
```diff
@@ -3,16 +3,13 @@
...
-  position: fixed;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
+  position: absolute;
+  top: env(safe-area-inset-top, 0px);
+  right: env(safe-area-inset-right, 0px);
+  bottom: env(safe-area-inset-bottom, 0px);
+  left: env(safe-area-inset-left, 0px);
```
akdetrick, line 10: The `env()` values are part of the viewport API. The `env()` function in CSS will fall back to the second argument if the first is undefined.

`safe-area-*` values correspond to the area of a mobile browser window **not** covered by chrome (e.g. keyboard)

rogerioshieh (commented):
Definitely better, but I still see the footer being covered by the search bar. I am not sure why this is happening, could it be because of the styling in the footer itself? It could also be something related to storybook, since it works when the tabs are rendered.


https://github.com/user-attachments/assets/794eaa82-9eb3-4227-aaad-f00dad791945

akdetrick:
@rogerioshieh Thanks for taking a look!

Have you tried it with "open canvas in new tab?"
Screenshot 2025-01-02 at 5 13 36 PM

This looks to me like a storybook-specific issue. Clicking on an individual story in the sidebar and using the open in new tab button will give you a full page story without an iframe. Example: https://60620d422ffdf100216415b2-djliidmeji.chromatic.com/iframe.html?args=&id=components-dialog--focus-management&viewMode=story

akdetrick:
Beta version published for testing against `banking`:

`4.2.0-beta.0`

akdetrick:
Looks good in `banking` (beta `4.2.0-beta.1`)

Screenshot 2025-01-03 at 11 13 39 AM

rogerioshieh (approved):
Ah, good call, I tried it and it works when opening in a new tab! Code looks great too 🚀

PR #1474, opened 2024-12-31
fix(Tabs): only add aria attr when panels are present

akdetrick:
Closes NDS-794

When Panels are included in a `Tabs` component, we link the tabs up with the panels using `role="tab"` and `aria-controls`.

When Panels are omitted, the tabs are basically just a list of links/buttons, so we do not include the `tab` role.

Without the `tab` role, the `aria-selected` attribute is invalid. This PR ensures that the `aria-selected` attribute is only present when panels are present.

### DIFF
--- a/src/Tabs/index.test.js
+++ b/src/Tabs/index.test.js
@@ -5,6 +5,7 @@ import Tabs from "./";
 const TAB_NAMES = ["Tab One", "Tab Two", "Tab Three"];
 const TAB_IDS = ["uno", "dos", "tres"];
 const PANEL_CONTENTS = ["Panel One", "Panel Two", "Panel Three"];
+const SELECTED_CLASS = "nds-tabs-tabItem--selected";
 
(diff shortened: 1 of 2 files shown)


src/Tabs/index.test.js
```diff
@@ -191,6 +192,7 @@ describe("Tabs", () => {
     [firstTab, secondTab, thirdTab].forEach((tab) => {
       expect(tab.parentElement).not.toHaveAttribute("role", "tab");
       expect(tab.parentElement).not.toHaveAttribute("aria-controls");
+      expect(tab.parentElement).not.toHaveAttribute("aria-selected");
     });
   });
```
akdetrick, line 197: This assertion is to test that we are NOT applying some ARIA attributes when there are no panels. We should assert `aria-selected` is not present when there are no panels.

TMiller00 (approved):

PR #1476, opened 2025-01-02
fix(Drawer): improve controls accessibility

akdetrick:
closes NDS-675

## Prior to this change
The Drawer had some nice hot keys assigned to page through panels using up/down arrows. Unfortunately, this can conflict with key bindings in child components.

## This change

- Removed arrow key bindings
- Converted control divs to `button` elements for easier keybaord nav
- Added an aria `region` around the content area that can be announced when prev/next buttons are used
- Used aria attributes to link up the buttons with the panel they control

Screenshot 2025-01-02 at 6 15 22 PM

### DIFF
--- a/src/Drawer/index.js
+++ b/src/Drawer/index.js
@@ -30,6 +30,7 @@ const Drawer = ({
   footer,
   testId,
 }) => {
+  const panelId = `content-panel-${Date.now()}`;
   const shimRef = useRef(null);
   const navRef = useRef(null);
 
@@ -50,22 +51,6 @@ const Drawer = ({
   const handleKeyDown = ({ key }) => {
     if (key === "Escape") {
       onUserDismiss();
-    }
-    if (!showControls) return;
-    if (isHorizontal) {
-      if (key === "ArrowRight") {
-        onNext && onNext();
-      }
-      if (key === "ArrowLeft") {
-        onPrev && onPrev();
-      }
-    } else {
-      if (key === "ArrowDown") {
-        onNext && onNext();
-      }
-      if (key === "ArrowUp") {
-        onPrev && onPrev();
-      }
     }
   };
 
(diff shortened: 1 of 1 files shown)


src/Drawer/index.js
```diff
@@ -99,12 +84,13 @@ const Drawer = ({
     >
       <div className={`navigation-container--${position}`}>
         {showClose && (
-          <div
-            className={`navigation-button navigation-button--${position} alignChild--center--center`}
+          <button
+            className={`button--reset navigation-button navigation-button--${position} alignChild--center--center`}
```
akdetrick, line 88: the humble `button--reset` class does a lot of the annoying button restyling work for us, making it easier to use real `button` elements instead of divs for custom controls like this

src/Drawer/index.js
```diff
@@ -128,30 +114,34 @@ const Drawer = ({
                 },
               ])}
               onClick={isHorizontal ? onNext : onPrev}
+              aria-controls={panelId}
+              aria-label={isHorizontal ? "Next" : "Previous"}
```
akdetrick, line 118: I'm pretty sure we can use some flex direction magic to solve for the `isHorizontal` checks, but it seems way out of scope for this one PR

src/Drawer/index.js
```diff
@@ -222,7 +225,13 @@ const Drawer = ({
...
       />
       {navigationContainerJSX}
-      {childrenJSX}
+      {!showControls ? (
+        childrenJSX
+      ) : (
+        <div aria-live="polite" role="region" id={panelId}>
+          {childrenJSX}
+        </div>
+      )}
```
akdetrick, line 234: It's either just some flat content, or dynamic region with controls

josh-narmi (approved):

PR #1478, opened 2025-01-03
fix(Select): allow actions when using categories

akdetrick:
Closes NDS-809

## The bug
When a `Select` with `Select.Category` components present also included a `Select.Action`, the action would create its own empty category.

Example:
image

## The fix
We now extract children a little differently, treating actions (`Select.Action`) as a different set than selectable options (`Select.Item`). 

Separating these out gives us full control over how each renders, which allows us to ensure the actions always render in a flat list at the bottom of the options list.

`Select.Action` and `Select.Children` are merged into the `items` array that gets passed to the Downshift hook. Downshift uses the `items` array by index to keep track of hover/arrow/selection states.

## Testing
In the storybook preview build, see the "Flat Categories" story of `Select`, which now contains an action.

### DIFF
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -129,13 +129,12 @@ const Select = ({
   testId,
 }) => {
   let items = []; // List of all item types to pass to downshift state management
-  let options = []; // All Select.Item options
   let categories = []; // Categories extracted from Select.Category children
-  let actions = []; // All Select.Action items
+  const options = React.Children.toArray(children).filter(
+    (item) => !isAction(item),
+  ); // All Select.Item options
+  const actions = React.Children.toArray(children).filter(isAction); // All Select.Action items
   const [userInput, setUserInput] = useState(""); // most recent val the user typed while focused on this input
-
-  options = React.Children.toArray(children).filter((item) => !isAction(item));
-  actions = React.Children.toArray(children).filter(isAction);
 
   // If categories are being used, extract items from categories
   if (options.some(({ type }) => type.displayName === "Select.Category")) {
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -58,9 +58,16 @@ export const getItemByValue = (value, items) => {
  * @param {array} items downshift index `items`
  * @returns {Number} index of item
  */
-export const getItemIndex = ({ props }, items) => {
-  const values = items.map(({ props }) => props.value);
-  return values.indexOf(props.value);
+export const getItemIndex = (item, items) => {
+  let result = 0;
+  if (isAction(item)) {
+    result = items
+      .map(({ props }) => props.onSelect)
+      .indexOf(item.props.onSelect);
+  } else {
+    result = items.map(({ props }) => props.value).indexOf(item.props.value);
+  }
+  return result;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/Select/index.js
```diff
@@ -58,9 +58,16 @@ export const getItemByValue = (value, items) => {
...
+  let result = 0;
+  if (isAction(item)) {
+    result = items
+      .map(({ props }) => props.onSelect)
+      .indexOf(item.props.onSelect);
+  } else {
+    result = items.map(({ props }) => props.value).indexOf(item.props.value);
+  }
+  return result;
 };
```
akdetrick, line 71: These changes exposed a bug in `getItemIndex` where it fails on an `Select.Action` subcomponent.

`Select.Item` can be found by their `value` prop. `Select.Action` should be found by their `onSelect`.

src/Select/index.js
```diff
@@ -86,7 +93,7 @@ export const isHighlightedInCategory = (
  * @returns {Boolean} if the selected item is in the given category children
  */
 export const isSelectedItemInCategory = (selectedItem, categoryChildren) => {
-  if (!selectedItem) return false;
+  if (!selectedItem || isAction(selectedItem)) return false;
```
akdetrick, line 96: Actions will _always_ be extracted from categories.

src/Select/index.js
```diff
@@ -58,9 +58,16 @@ export const getItemByValue = (value, items) => {
...
+export const getItemIndex = (item, items) => {
+  let result = 0;
+  if (isAction(item)) {
+    result = items
+      .map(({ props }) => props.onSelect)
+      .indexOf(item.props.onSelect);
+  } else {
+    result = items.map(({ props }) => props.value).indexOf(item.props.value);
+  }
+  return result;
```
graphicsforge, line 70: Do you think this would be a touch more legible if expressed this way?
```suggestion
  // find by matching onSelect callback
  if (item?.props?.onSelect) {
    return items
      .map(({ props }) => props.onSelect)
      .indexOf(item.props.onSelect);
  } 
  // find by matching value
  return items.map(({ props }) => props.value).indexOf(item.props.value);
```
return should be -1 if any indexOf doesn't find a match, not 0... but `let result = 0;` should always get overridden anyways, right?

src/Select/index.js
```diff
@@ -121,18 +128,22 @@ const Select = ({
...
-  let categories = [];
+  let items = []; // List of all item types to pass to downshift state management
+  let options = []; // All Select.Item options
+  let categories = []; // Categories extracted from Select.Category children
+  let actions = []; // All Select.Action items
   const [userInput, setUserInput] = useState(""); // most recent val the user typed while focused on this input
 
-  const allChildren = React.Children.toArray(children);
+  options = React.Children.toArray(children).filter((item) => !isAction(item));
+  actions = React.Children.toArray(children).filter(isAction);
```
graphicsforge, line 138: Ah, so we can always put the actions at the bottom? Sounds good, that'll help us enforce the design!
How would you feel about making these two `const` instead of `let` to indicate that there's no further reassignment? 🤔

graphicsforge (approved):
oh yes, we want to pull out actions from categories. This change makes sense to me 👍

src/Select/index.js
akdetrick: I do like your style better, but I initialize `result = 0` because downshift needs the index to be in range.

A result of `-1` should never happen, but if for some reason there's a bug and we can't find an item index, downshift state should just use the first item (index `0`)

src/Select/index.js
akdetrick: Yeah, I can just directly assign these where `items` is initialized... Going to make this change, thanks!

PR #1485, opened 2025-01-07
feat(RadioButtons): add kind "row-start" to radio buttons

trentslutzky:
Adds "row-start" to the kind prop for Radio buttons
Removes the justify: space-between and adds a gap 

image

### DIFF
--- a/src/RadioButtons/index.scss
+++ b/src/RadioButtons/index.scss
@@ -84,6 +84,58 @@
   }
 }
 
+// row-start variant of radiobuttons
+.nds-radiobuttons--row-start {
+  display: flex;
+  gap: var(--space-m);
+
+  .nds-radiobuttons-option {
+    padding-left: rem(28px);
+
+    .nds-radio {
+      position: absolute;
+      top: 0;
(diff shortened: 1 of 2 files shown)


src/RadioButtons/index.scss
```diff
@@ -84,6 +84,58 @@
   }
 }
 
+// row-start variant of radiobuttons
+.nds-radiobuttons--row-start {
```
akdetrick, line 88: A lot of this code duplicates what `row` has, BUT keeping the styles entirely separate keeps them uncoupled, which is better for the long term.

This was the right call

akdetrick (changes requested):
In NDS, the PR title doesn't matter, but the commit messages do.

We'll want at least one commit in this PR with a `feat(RadioButtons): ` message

akdetrick (approved):

akdetrick (changes requested):
Sorry, I meant the actual commit messages. `semantic-release` analyzes the commit messages to determine the release

akdetrick (approved):

PR #1486, opened 2025-01-07
fix(Select): increase cateogry header style specificity

akdetrick:
closes NDS-849

## The bug
Category headers render as `h4` html elements. There are some places in `banking` that still rely on the `nds-typography` class, which includes `font-size` and `font-weight` rules for all heading tags, causing awkwardly big category headers:
Screenshot 2025-01-07 at 5 03 59 PM

## The fix

- Increase specificity of Select category label styles
- Remove vertical padding from collapsible categories

Screenshot 2025-01-07 at 5 10 16 PM


ethan-hogen (approved):
this makes sense to me!

PR #1487, opened 2025-01-08
fix(AutocompleteModal): make AutocompleteModal height adjust to small amount of items

xian0831:
When the number of auto complete items are not enough to take up all the vertical space, it will show white background. This behavior is caused by using fixed height on `AutocompleteModal`. 

### With fixed height 
Screenshot 2025-01-08 at 11 10 50 AM

### With height adjusts to the content
Screenshot 2025-01-08 at 11 11 29 AM

Ticket: https://linear.app/narmi/issue/NDS-850/[nds]-make-autocompletemodal-height-adjustable

### DIFF
--- a/src/AutocompleteModal/index.scss
+++ b/src/AutocompleteModal/index.scss
@@ -10,7 +10,8 @@
 .nds-autocompleteModal-positionedEl {
   min-width: rem(248px);
   max-width: rem(352px);
-  height: rem(312px);
+  height: auto;
+  max-height: rem(312px);
   display: flex;
   flex-direction: column;
 


src/AutocompleteModal/index.scss
```diff
@@ -10,7 +10,8 @@
 .nds-autocompleteModal-positionedEl {
   min-width: rem(248px);
   max-width: rem(352px);
-  height: rem(312px);
+  height: auto;
+  max-height: rem(312px);
```
xian0831, line 14: `max-height` make sure this change doesn't impact existing usage of AutocompleteModal

akdetrick: love it

akdetrick (approved):
Failing check is likely our chromatic limit. 

Code changes look good 👍

PR #1491, opened 2025-01-10
refactor(Dialog): convert Dialog to typescript

akdetrick:
Closes NDS-874

Converts `Dialog` to typescript.


src/index.ts
```diff
@@ -56,7 +56,6 @@ declare const formatDate;
 
 export * from "./types/Icon.types";
 export {
-  // typed
```
akdetrick, line 59: The distinction between typed and untyped components in this `export` doesn't really matter as long as an export is present for each component.

Simplifying this

xian0831 (approved):

PR #1497, opened 2025-01-10
refactor(Tabs): convert Tabs to TypeScript

xian0831:
https://linear.app/narmi/issue/NDS-887/[nds-typescript]-convert-tabs-to-typescript

### DIFF
--- a/src/Tabs/Arrow.tsx
+++ b/src/Tabs/Arrow.tsx
@@ -1,10 +1,15 @@
-import PropTypes from "prop-types";
 import React, { useContext, useRef } from "react";
 import { CSSTransition } from "react-transition-group";
 import IconButton from "../IconButton";
 import TabsContext from "./context";
 
-const Arrow = ({ direction, onClick, show }) => {
+interface ArrowProps {
+  direction: string;
+  onClick?: () => void;
+  show?: boolean;
+}
+
+const Arrow = ({ direction, onClick, show }: ArrowProps) => {
   const { isResponsive } = useContext(TabsContext);
   const arrowRef = useRef();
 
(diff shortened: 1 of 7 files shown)


src/Tabs/Arrow.tsx
```diff
@@ -31,14 +36,4 @@ const Arrow = ({ direction, onClick, show }) => {
   );
 };
 
-Arrow.propTypes = {
-  /** Direction of the arrow button */
-  direction: PropTypes.string.isRequired,
-  onClick: PropTypes.func,
-  show: PropTypes.bool,
-  isResponsive: PropTypes.bool,
```
xian0831, line 39: `isResponsive ` is not passed in as a prop. Remove it from the type definition.

src/Tabs/TabsList.tsx
```diff
@@ -128,11 +137,11 @@ const TabsList = ({ children, xPadding = "none" }) => {
     <div className={isResponsive ? "display-flex" : ""}>
       <Arrow direction="left" onClick={onLeftClick} show={showLeftArrow} />
       <ul
-        ref={tabsListRef}
+        ref={tabsListRef as LegacyRef<HTMLUListElement>}
```
xian0831, line 140: Maybe there is a more proper way to do this. But it solves the type issue.

src/Tabs/TabsList.tsx
```diff
@@ -128,11 +137,11 @@ const TabsList = ({ children, xPadding = "none" }) => {
...
         role={hasPanels ? "tablist" : undefined}
         className={`nds-tabs-tabsList list--reset padding--x--${xPadding}`}
         onKeyDown={hasPanels ? handleKeyDown : noop}
-        tabIndex={hasPanels ? "0" : undefined}
+        tabIndex={hasPanels ? 0 : undefined}
```
xian0831, line 144: tabIndex takes number

src/Tabs/TabsList.tsx
akdetrick: Just learned about this today, but it seems like the right way to go

akdetrick (approved):

PR #1503, opened 2025-01-18
Upgrade to Storybook Version 8

xian0831:
My initial motivation for upgrading SB to version 8 is that we are using a component pattern where some components are used as sub-components, such as `Select` and `Select.Item`   or `Combobox` and `Combobox.Item`. This patterns works fine but SB 7 removed the support for `subcomponent`.  This makes it extremely bad DX where developers will need to look into the source code to know correct props to use and how to use them on those sub-components.

## Fixed breaking changes from SB 7

1. Update React to 18 in **dev dependency.**
2. Update React Testing Library. Confirmed all test cases passed as expected.   
3. In MDX, `Story` no longer accept children, so I moved split the actual stories from MDX to separate `*.stories.js` files. MDX will focus on how to render stories properly using Canvas.
4. Fix package structure changes. [See details](https://storybook.js.org/docs/migration-guide#package-structure-changes)
5. ~Add [`storybook-addon-tag-badges`](https://github.com/Sidnioulz/storybook-addon-tag-badges?ref=storybookblog.ghost.io) addon to surface the new Tag features on the sidebar.~ I'm reaching out to the package author to fix peer dependency. Currently it is tied to SB 8.4 

#### Showing sub component properly

image

## Follow up items
- Since `argTypesRegex` is removed from SB 8. We will need to explicitly add arg to components that wants to show "Action" addon. I don't think this is a blocker given how often developers use this feature.
- Given that we are on the latest version of Storybook, which seems to added great support for React Native. It may be something worth look into by the platform team and the mobile team. https://storybook.js.org/blog/react-native-storybook-8/


.gitignore
```diff
@@ -2,3 +2,6 @@ node_modules
 dist
 .DS_Store
 storybook-static
+
+# IntelliJ IDEA
+.idea
```
xian0831, line 7: ignore config files from JetBrain IDEs

akdetrick (commented):
These changes look great! I really appreciate the contribution because it directly helps make a path for to achieve some of our UI Platform goals sooner than expected

Everything looks good so far; I want to give this a more careful review Wednesday after I merge an NDS feature blocking some accessibility work, then let's 🚢

xian0831:
> These changes look great! I really appreciate the contribution because it directly helps make a path for to achieve some of our UI Platform goals sooner than expected
> 
> Everything looks good so far; I want to give this a more careful review Wednesday after I merge an NDS feature blocking some accessibility work, then let's 🚢

Curious on those UI Platform goals. Can you share a little bit of details so I can gain some context?

tokens/src/color/color.mdx
```diff
@@ -1,32 +1,23 @@
...
+import { Meta, Story, Canvas } from "@storybook/blocks";
 import TokenTable, { toTokenRows } from "helpers/TokenTable";
+import * as ColorStories from "./color.stories.js";
 
 <Meta title="Design Tokens/Color" />
 
 # Color Tokens
 
-## Narmi Base Colors
+## Narmi Base ColorsO
```
akdetrick, line 9:
```suggestion
## Narmi Base Colors
```

akdetrick (approved):

PR #1510, opened 2025-01-27
feat(Radio): add single Radio component

akdetrick:
Closes NDS-922

Adds the `Radio` component, for rendering single radio inputs with full control over labels and markup between radios.

### DIFF
--- a/src/Radio/index.scss
+++ b/src/Radio/index.scss
@@ -1,0 +1,47 @@
+.nds-singleRadio {
+  position: relative;
+}
+
+.nds-singleRadio input {
+  position: absolute;
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
+}
+
+.nds-singleRadio-radio {
+  // outline of the custom radio control
+  position: relative;
+  height: rem(20px);
+  width: rem(20px);
+  background-color: var(--color-white);
+  border: 1px solid var(--color-lightGrey);
+  border-radius: 50%;
+
+  // initially hidden dot
+  &:after {
+    position: absolute;
+    display: none;
+    content: "";
+    top: 50%;
+    left: 50%;
+    width: rem(12px);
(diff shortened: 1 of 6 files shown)


src/Radio/index.scss
```diff
@@ -0,0 +1,47 @@
...
+.nds-singleRadio input {
+  position: absolute;
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
+}
```
akdetrick, line 14: This should probably be a utility class, but that's out of scope for this PR.

Basically this set of rules hides the actual `input` element from users who use their eyes to find the element, but allows it to remain visible for assistive tech.

If you set `display` to `none`, it will be hidden from assistive tech.

This is also why I place a `presentation` role on the div that paints the faux input.

src/Radio/index.scss
```diff
@@ -0,0 +1,47 @@
...
+
+.nds-singleRadio input:checked + .nds-singleRadio-radio,
+.nds-singleRadio:focus-within .nds-singleRadio-radio {
+  border: 1px solid var(--theme-primary);
+}
+
+.nds-singleRadio input:checked + .nds-singleRadio-radio:after {
+  display: block;
+}
```
akdetrick, line 47: Instead of using classcat and internal state to manage styling, `input:checked` and `.nds-singleRadio:focus-within` give us everything we need

blooies (approved):
looks good to me! thanks for adding!

src/Radio/index.stories.js
```diff
@@ -0,0 +1,67 @@
...
+      Repeats <strong>Weekly</strong>
+    </Radio>
+    <Radio name="frequency" value="monthly">
+      Repeats <strong>Monthly</strong>
+    </Radio>
+  </div>
+);
+
+export const FullyControlled = () => {
+  const SET_NAME = "frequency";
```
akdetrick, line 40:
```suggestion
  const SET_NAME = "freq";
```

PR #1511, opened 2025-01-27
feat(Combobox): add hint variant for heading items

akdetrick:
closes NDS-780

Adds `hint` variant for `Combobox.Heading` Items.

Screenshot 2025-01-27 at 4 54 39 PM

### DIFF
--- a/src/Combobox/ComboboxHeading.js
+++ b/src/Combobox/ComboboxHeading.js
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 
 const ComboboxHeading = ({ text, kind = "bold" }) => (
   <h5
-    className={`nds-combobox-list-heading--${kind} fontSize--default fontWeight--bold`}
+    className={`nds-combobox-list-heading--${kind}`}
   >
     {text}
   </h5>
--- a/src/Combobox/ComboboxHeading.js
+++ b/src/Combobox/ComboboxHeading.js
@@ -1,8 +1,10 @@
 import React from "react";
 import PropTypes from "prop-types";
 
-const ComboboxHeading = ({ text }) => (
-  <h5 className="nds-combobox-list-heading fontSize--default fontWeight--bold">
+const ComboboxHeading = ({ text, kind = "bold" }) => (
+  <h5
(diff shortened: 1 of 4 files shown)


src/Combobox/ComboboxHeading.js
```diff
@@ -1,8 +1,10 @@
 import React from "react";
 import PropTypes from "prop-types";
 
-const ComboboxHeading = ({ text }) => (
-  <h5 className="nds-combobox-list-heading fontSize--default fontWeight--bold">
+const ComboboxHeading = ({ text, kind = "bold" }) => (
```
akdetrick, line 4: `bold` represents the existing heading style. This is a non-breaking change.

src/Combobox/index.scss
```diff
@@ -97,3 +97,17 @@
     transform: translate(rem(1px), -50%);
   }
 }
+
+.nds-combobox-list-heading--bold {
```
akdetrick, line 101: I'm using `!important` on font properties in these rulesets because the legacy `.nds-typography` class may be used in `banking`, which requires a higher specificity to override default styles on heading tags.

src/Combobox/ComboboxHeading.js
```diff
@@ -1,8 +1,10 @@
...
+  <h5
+    className={`nds-combobox-list-heading--${kind} fontSize--default fontWeight--bold`}
```
xian0831, line 6: Given font size and weight will be controlled by `nds-combobox-list-heading--${kind}`, should we just remove `fontSize--default fontWeight--bold`?

akdetrick: There's an old `nds-typography` class that will style child headings a certain way and this ensures the weight and size regardless of parent clasess

akdetrick: Wait, I think I understand now- taking a closer look

akdetrick, line 6:
```suggestion
    className={`nds-combobox-list-heading--${kind}`}
```

xian0831 (approved):

PR #1516, opened 2025-01-30
fix: prevent TS errors in consumers

akdetrick:
Unblocks NDS upgrades due to type errors:

e.g. https://github.com/narmi/banking/actions/runs/13059724173/job/36439440220?pr=56395

### DIFF
--- a/src/Button/index.tsx
+++ b/src/Button/index.tsx
@@ -7,11 +7,11 @@ import type { IconName } from "../types/Icon.types";
 
 interface ButtonProps {
   /** Renders the button label */
-  label: string;
+  label?: string; // must be optional until `children` is removed
   /** style of button to render */
   kind?: "primary" | "secondary" | "tonal" | "negative" | "menu" | "plain";
   /** Click callback, with event object passed as argument */
-  onClick?: (e: React.MouseEvent) => void;
+  onClick?: (e) => void;
   /**
    * The html element to render as the root node of `Button`.
    *
--- a/src/ContentCard/index.tsx
+++ b/src/ContentCard/index.tsx
@@ -34,7 +34,7 @@ interface ContentCardProps {
    * Amount of border radius to add on all sides of card.
    */
   radiusSize?: "s" | "m" | "l";
-  onClick?: () => void;
+  onClick?: (e: React.MouseEvent | React.KeyboardEvent) => void;
   /**
    * Only applicable for `toggle` type.
    * Renders card in visually selected state with appropriate attributes.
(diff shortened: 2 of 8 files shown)


src/Button/index.tsx
```diff
@@ -7,11 +7,11 @@ import type { IconName } from "../types/Icon.types";
 
 interface ButtonProps {
   /** Renders the button label */
-  label: string;
+  label?: string; // must be optional until `children` is removed
```
akdetrick, line 10: We still support passing `children` as a label for legacy usage

src/ContentCard/index.tsx
```diff
@@ -34,7 +34,7 @@ interface ContentCardProps {
    * Amount of border radius to add on all sides of card.
    */
   radiusSize?: "s" | "m" | "l";
-  onClick?: () => void;
+  onClick?: (e?: React.MouseEvent) => void;
```
akdetrick, line 37: We do NOT pass an event object when calling the passed onClick function when handling key events, but it looks like the event is passed on mouse click events.

src/RadioButtons/index.tsx
```diff
@@ -20,7 +14,7 @@ interface RadioButtonsProps {
    *   | string;
    * ```
    * */
-  options: Record<string, OptionType>;
+  options?: object;
```
akdetrick, line 17: I would like to revisit this, but in the meantime this looser type prevents type errors throughout the staff app

src/Dialog/index.tsx
```diff
@@ -29,7 +29,7 @@ export interface DialogProps {
   /** Scrollable contents of the Dialog */
   children: React.ReactNode;
   /** Heading in the top of the Dialog */
-  title: string;
+  title?: string;
```
xian0831, line 32: We need handle the situation when title is undefined. I think it is controlled by `isBanner`


```
{isBanner ? (
                <div className="margin--y--s">
                  <div
                    className="padding--y--xxs"
                    id="aria-dialog-label"
                    style={{ textAlign: "center" }}
                  >
                    {title}
                  </div>
                  {closeButtonJSX}
                </div>
              ) : (
                <>
                  <h4 id="aria-dialog-label">{title}</h4>
                  {closeButtonJSX}
                </>
              )}
```

src/ContentCard/index.tsx
akdetrick: I like that.

src/Dialog/index.tsx
akdetrick: Set a default value instead

xian0831 (approved):

PR #1525, opened 2025-02-06
fix(Dialog): position shim relative to viewport instead of document

akdetrick:
closes NDS-978

## The bug
We recently made a fix for iOS Safari to make Dialog positioning mobile keyboard aware, but we introduced a regression.
The Dialog shim was positioning itself relative to the `_document_` instead of the viewport. This causes layout issues when users are scrolled down a tall document element. as we observed in NDS-977.

## The fix
We always intended the DIalog to position itself relative to the viewport, so I added a root element that is positioned `fixed`.

Any `absolute` children of the root element will use the _nearest positioned parent_ instead of the document as its positioning reference.F

## Does it still work in iOS?
Yes.

Feb-06-2025 16-41-43

### DIFF
--- a/src/Dialog/index.scss
+++ b/src/Dialog/index.scss
@@ -2,6 +2,11 @@
   --overflow-gradient-height: 20px;
 }
 
+.nds-dialog-root {
+  position: fixed;
+  inset: 0;
+}
+
 .nds-shim--dark {
   position: absolute;
   top: env(safe-area-inset-top, 0px);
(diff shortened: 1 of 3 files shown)


src/Dialog/index.scss
```diff
@@ -66,6 +71,7 @@
   border-top-left-radius: var(--border-radius-m);
 
   max-height: 96%; // relative to parent, .nds-dialog-focuslock
+  min-width: 100dvw;
```
akdetrick, line 74: This fixes an unrelated minor bug I found when testing. It's quite an edge case, but if a Dialog doesn't have very much content in it (too little to wrap a single line of text), it won't take up the whole screen width at mobile sizes.

This is a safe change because we already override `min-width` for desktop sizes.

tlackemann (approved):

PR #1532, opened 2025-02-11
fix(drawer): add a z-index and position

kevmo314:
Adds a `z-index` and `position: relative;` to create a new stacking context. This gives the page 100 z-indexes to use for stacking without interfering with the drawer.

### DIFF
--- a/src/Drawer/index.js
+++ b/src/Drawer/index.js
@@ -235,7 +235,7 @@ const Drawer = ({
     </div>
   );
 
-  function renderDdrawerInOutlet() {
+  function renderDrawerInOutlet() {
     // create a single outlet
     if (!document.getElementById("outlet")) {
       const outlet = document.createElement("div");
@@ -249,7 +249,7 @@ const Drawer = ({
     return null;
   }
 
-  return <>{document && renderDdrawerInOutlet()}</>;
+  return <>{document && renderDrawerInOutlet()}</>;
 };
 
 Drawer.propTypes = {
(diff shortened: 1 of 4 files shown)


src/Drawer/index.scss
```diff
@@ -2,6 +2,8 @@
   --transition-speed-fast: 0.4s;
   --transition-speed-medium: 0.7s;
   --transition-speed-slow: 1s;
+  z-index: 100;
+  position: relative;
```
akdetrick, line 6: @kevmo314 Does this work when a Dialog, Popover, or Select is in drawer content?

xian0831: @adamcadaver @kevmo314 I tested locally. They worked as expected 

Screenshot 2025-02-18 at 2 00 50 PM
Screenshot 2025-02-18 at 1 59 32 PM
Screenshot 2025-02-18 at 1 58 01 PM

src/Drawer/index.scss
```diff
@@ -2,6 +2,7 @@
   --transition-speed-fast: 0.4s;
   --transition-speed-medium: 0.7s;
   --transition-speed-slow: 1s;
+  position: fixed;
```
xian0831, line 5: Since we can remove fixed position from Dialog, we have to change the position to fixed for drawer to make sure stacking works properly

src/Drawer/index.js
```diff
@@ -235,10 +235,21 @@ const Drawer = ({
     </div>
   );
 
+  function renderDdrawerInOutlet() {
```
kevmo314, line 238:
```suggestion
  function renderDrawerInOutlet() {
```

akdetrick (approved):
Verified working in storybook! Thanks for the fix

PR #1533, opened 2025-02-11
feat(TruncatedAccount): add option for inline rendering

akdetrick:
closes NDS-1053

Adds `isInline` prop to allow inline display of `TruncatedAccount`


aundus (approved):
Looks great ty!

PR #1534, opened 2025-02-13
Refactor/nds 812 bump downshift

akdetrick:
Closes NDS-812

- Bumps downshift to v9
- Handles breaking changes

#### Affected components:
- Select
- Combobox
- MultiSelect
- AutocompleteModal

### DIFF
--- a/src/AutocompleteModal/AutoComplete.tsx
+++ b/src/AutocompleteModal/AutoComplete.tsx
@@ -48,7 +48,6 @@ const AutoComplete = ({
   const {
     selectedItem,
     highlightedIndex,
-    getComboboxProps,
     getMenuProps,
     getInputProps,
     getItemProps,
@@ -74,10 +73,7 @@ const AutoComplete = ({
   });
 
   return (
-    <div
-      className={cc(["nds-autocompleteModal-autocomplete", "rounded--all"])}
-      {...getComboboxProps()}
-    >
+    <div className={cc(["nds-autocompleteModal-autocomplete", "rounded--all"])}>
       <div className="nds-autocompleteModal-input">
         <TextInput label={label} {...getInputProps()} />
       </div>
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -179,7 +179,6 @@ const Combobox = ({
     isOpen,
     selectedItem,
     getInputProps,
-    getComboboxProps,
     getMenuProps,
     getItemProps,
     highlightedIndex,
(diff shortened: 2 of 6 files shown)


src/AutocompleteModal/AutoComplete.tsx
```diff
@@ -74,10 +73,7 @@ const AutoComplete = ({
   });
 
   return (
-    <div
-      className={cc(["nds-autocompleteModal-autocomplete", "rounded--all"])}
-      {...getComboboxProps()}
-    >
```
akdetrick, line 80: This is the biggest change. The ARIA spec for the combobox pattern changed from ARIA 1.1 to 1.2

The recommended pattern is now to only apply the aria attributes to the input itself rather than a common parent between the input and the menu.

Downshift removed `getComboboxProps` as a result because it's no longer needed.

src/Combobox/index.js
```diff
@@ -401,8 +400,8 @@ const Combobox = ({
     <>
       <div
         className={cc(["nds-combobox", { "nds-combobox--active": isOpen }])}
-        {...getComboboxProps(triggerProps)}
         data-testid={testId}
+        {...triggerProps}
```
akdetrick, line 404: `triggerProps` comes from our modal positioning engine hook, `useLayer`. 

We still need these props spread onto the parent because `useLayer` needs it as a positioning reference.

src/Select/index.js
```diff
@@ -177,10 +177,7 @@ const Select = ({
       const { type, changes } = actionAndChanges;
       let isOpen = changes.isOpen;
 
-      if (
-        type === useSelect.stateChangeTypes.MenuKeyDownCharacter ||
-        type === useSelect.stateChangeTypes.ToggleButtonKeyDownCharacter
-      ) {
+      if (type === useSelect.stateChangeTypes.ToggleButtonKeyDownCharacter) {
```
akdetrick, line 180: `MenuKeyDownCharacter` was removed in v8

src/Select/index.js
```diff
@@ -213,7 +210,8 @@ const Select = ({
   } = useSelect(downshiftOpts);
 
   const hasCategories = categories.length > 0;
-  const hasSelectedItem = selectedItem !== undefined && selectedItem.props;
+  console.dir(selectedItem);
```
akdetrick, line 213:
```suggestion
```

src/Select/index.js
```diff
@@ -213,7 +210,7 @@ const Select = ({
   } = useSelect(downshiftOpts);
 
   const hasCategories = categories.length > 0;
-  const hasSelectedItem = selectedItem !== undefined && selectedItem.props;
+  const hasSelectedItem = selectedItem !== null && selectedItem.props;
```
akdetrick, line 213: As of downshift v7, `selectedItem` initializes to `null`.

src/Select/index.js
akdetrick, line 213:
```suggestion
```

josh-narmi (approved):

PR #1543, opened 2025-02-20
fix(collapsibleCard): title color should be theme primary

rogerioshieh:
https://linear.app/narmi/issue/DEP-2471/update-colors-in-staff-led

The difference is subtle with the Moss green set as primary theme color:

Now:
image

Previously:
image


rogerioshieh:
There's is a `renderTitle` that would give the expected output.

PR #1544, opened 2025-02-21
fix(Drawer,Dialog): apply position and z-index on portal outlet

xian0831:
Previous PR (https://github.com/narmi/design_system/pull/1532) address the z-index issue among a few NDS components. 
This PR includes z-index and position changes to make sure portal outlet element is properly overlay on top of other non-NDS components in the app when they use z-index. 

The following is an example of the issue I'm trying to address. The action bar with header "Applications" has the a z-index of 2.

image


akdetrick (approved):
makes sense!

PR #1545, opened 2025-02-21
fix(Dialog): prevent focus trap el from blocking shim clicks

akdetrick:
Closes NDS-1100

As Wenting helpfully pointed out, there's a specific area of the screen where you click the shim and nothing happens.

dialog_inconsistent_dismissal_releases0
Screenshot 2025-02-21 at 9 42 19 AM

As it turns out the `FocusTrap` renders a div that has an `auto` width at screen sizes larger than `s`. 

This PR fixes this issue so that the focus trap element "shrink wraps" to the dialog area so it doesn't cover up the shim and intercept shim clicks.

### Mobile Safari is still good 👍 
Screenshot 2025-02-21 at 2 56 00 PM
Screenshot 2025-02-21 at 2 56 06 PM

### DIFF
--- a/src/Dialog/index.tsx
+++ b/src/Dialog/index.tsx
@@ -121,8 +121,8 @@ const Dialog = ({
   // the shim has events for mouse users only; does not require a role
   /* eslint-disable jsx-a11y/no-static-element-interactions,jsx-a11y/click-events-have-key-events */
   const dialogJSX = (
-    <div className="nds-dialog-root">
-      <CSSTransition timeout={1} classNames="nds-dialog-transition" appear in>
+    <CSSTransition timeout={1} classNames="nds-dialog-transition" appear in>
+      <div className="nds-dialog-root">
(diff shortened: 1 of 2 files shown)


src/Dialog/index.tsx
```diff
@@ -121,8 +121,8 @@ const Dialog = ({
   // the shim has events for mouse users only; does not require a role
   /* eslint-disable jsx-a11y/no-static-element-interactions,jsx-a11y/click-events-have-key-events */
   const dialogJSX = (
-    <div className="nds-dialog-root">
-      <CSSTransition timeout={1} classNames="nds-dialog-transition" appear in>
+    <CSSTransition timeout={1} classNames="nds-dialog-transition" appear in>
```
akdetrick, line 124: I moved the CSSTransition to the outermost element here to restore the transition. It actually transitions now.

josh-narmi (approved):

Issue #1551, opened 2025-02-25
[Checkbox]: checkbox element should be vertically centered if label is multiline


agreatscott:
### Component/Module

Checkbox

### Screenshots (if applicable)

![Image](https://github.com/user-attachments/assets/dfb5a738-b7c3-40cc-8c11-e95ca203c6bf)

### Reproduction Steps

Use `checkbox` with either a very long label string or adjust container to be small width

### Expected Result

checkbox element itself is centered vertically

### Actual Result

checkbox element itself is aligned to the top of the container

agreatscott:
my mistake - checked with design and this is not a bug, current display is fine

PR #1552, opened 2025-02-25
fix(checkbox): align checkbox itself vertically when label is multiline

agreatscott:
resolves https://github.com/narmi/design_system/issues/1551

This bug has been bothering me and is a fairly straightforward fix. I'll double check with design to make sure this is actually  desired. 

image


agreatscott:
Confirmed with design that this is not a bug, so closing

PR #1556, opened 2025-02-28
feat(Alert): expose more padding sizes for Alert

xian0831:
In the Organization Profile project for the Staff app, we need to use the padding size "s", but it is not valid from TypeScript point of view. 

I think it is too inconvenient for developers to updating NDS when they need a new size.  We are already supporting all padding sizes, this PR is just going to update the doc and type definitions.


akdetrick (approved):
`paddingSize` already has a default value; changes look good

PR #1562, opened 2025-03-03
fix: use correct ProgressBar import

TMiller00:
In the `index.js`, `ProgressBar` was being imported from `./Pagination`. This PR updates the import to occur from `./ProgressBar`


akdetrick (approved):
thanks!

PR #1563, opened 2025-03-03
fix(Combobox): blur input when selecting actions

akdetrick:
Closes NDS-1134

The input should blur when an action is selected in the combobox dropdown, but that wasn't happening.  We missed this bug because `Dialog` correctly has a focus trap and steals focus from the input. When using a `Combobox.Action` to open a `Drawer`, the drawer does not steal focus and the document.activeElement is still the combobox input behind the shim.

This PR fixes this by ensuring the input is always blurred when a `Combobox.Action` is selected.

- Capture input ref in downshift's prop getter
- When selection changes to an action, blur the input in addition to closing the menu

### DIFF
--- a/src/Combobox/index.stories.js
+++ b/src/Combobox/index.stories.js
@@ -4,6 +4,7 @@ import Combobox, { VALID_ICON_NAMES } from "./";
 import ComboboxItem from "./ComboboxItem";
 import ComboboxHeading from "./ComboboxHeading";
 import Dialog from "../Dialog";
+import Drawer from "../Drawer";
 import { options_states } from "./util";
 
 const Template = (args) => <Combobox {...args} />;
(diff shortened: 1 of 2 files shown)


src/Combobox/index.stories.js
```diff
@@ -237,14 +238,17 @@ CustomFiltering.parameters = {
   },
 };
 
-export const WithActions = (args) => {
+export const WithActions = () => {
   const [isOpen, setIsOpen] = useState(false);
+  const [inputValue, setInputValue] = useState("");
```
akdetrick, line 243: Change example to a controlled component, which is the norm for us and expected by `ContextForm` and DBBL form adapters.

src/Combobox/index.stories.js
```diff
@@ -253,13 +257,9 @@ export const WithActions = (args) => {
           label="Do Action"
         />
       </Combobox>
-      <Dialog
-        title="You selected an action"
-        isOpen={isOpen}
-        onUserDismiss={() => setIsOpen(false)}
-      >
+      <Drawer isOpen={isOpen} onUserDismiss={() => setIsOpen(false)}>
```
akdetrick, line 260: `Drawer` is a better pick for the actions story because it doesn't steal focus by default.

blooies (approved):
lgtm!!

PR #1573, opened 2025-03-24
feat(ProgressBar): update component to match new design sizes and styles

akdetrick:
Closes NDS-935

Updates `ProgressBar` to match design standards. As of 2025, no component in `banking` actually uses this component and changes are non-breaking (third party consumers will just get an aesthetic update).

- Update styling
- Remove CSS transition that wasn't working
- Add SMIL animation node to SVG line to manage animating between values
- Convert to typescript
- Add `size` variants


src/ProgressBar/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+      }
+    }
+  }, [percentComplete]);
+
+  return (
+    <div
+      className={`nds-progressbar nds-progressbar--${size}`}
+      style={{ width: `${totalWidth}` }}
+    >
+      {percentComplete > 0 && (
```
akdetrick, line 45: The rounded line cap in SVG is drawn from the _end_ of the line, so with an `x2` value of `0`, I was seeing a little round thing peek out. This prevents the progress line from being drawn if there's nothing to draw.

src/ProgressBar/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const animateRef = useRef<SVGAnimateElement>(null);
+
+  useLayoutEffect(() => {
+    const animFrom = percentComplete === 0 ? percentComplete : currentPct;
+    if (lineRef.current) {
+      if (animateRef.current) {
+        animateRef.current.setAttribute("from", `${animFrom}%`);
+        animateRef.current.setAttribute("to", `${percentComplete}%`);
+        animateRef.current.beginElement();
+        setCurrentPct(percentComplete);
```
akdetrick, line 35: Triggers animation on the `line` element and stores the current `percentComplete` for the next render

src/ProgressBar/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <svg xmlns="http://www.w3.org/2000/svg">
+          <line
+            ref={lineRef}
+            x1="0%"
+            y1="50%"
+            x2={`${currentPct}%`}
+            y2="50%"
+            strokeLinecap="round"
+          >
+            <animate
```
akdetrick, line 68: <https://developer.mozilla.org/en-US/docs/Web/SVG/Guides/SVG_animation_with_SMIL>

TMiller00 (approved):

PR #1574, opened 2025-03-26
fix(TextInput): show char counter at 0 chars

akdetrick:
Closes NDS-1143

Per design, change `TextInput` char counter behavior to _always_ show the counter, even at 0.


XihaoLuo (approved):
Makes sense.

PR #1576, opened 2025-03-27
feat(MultiSelect): Adds the new customizable summary mode and clear-all

JohnathanWeisner:
resolves https://linear.app/narmi/issue/CDB-748/[web]-add-summary-mode-to-multiselect-add-clear-all-button
This adds a summary mode to the MultiSelect component.
Fixes a bug around controlled state updates.
Adds a clear all button.

### DIFF
--- a/src/MultiSelect/index.js
+++ b/src/MultiSelect/index.js
@@ -33,14 +33,46 @@ const getSelectedItems = (values, items) =>
   items.filter((item) => values.includes(item.props.value));
 
 /**
- * Get new list of `MenuSelect.item` components that are selected after
- * user dismisses a token.
- * @param {Array} newTokenLabels
- * @param {Array} selectedItems
- * @returns Array
+ * Default summary formatter function.
+ *
+ * Receives an object with:
+ *   - selectedItems: Array of currently selected item React elements.
+ *   - getSelectedItemProps: Helper function for token props.
+ *   - removeSelectedItem: Function to remove a token.
+ *   - disabled: Flag indicating if the component is disabled.
+ *   - label: Placeholder text when no items are selected.
+ *
+ * If no items are selected, returns the label.
+ * Otherwise, returns a container with FieldToken components.
+ * Each token displays the tokenLabel (if provided) or falls back to the string value.
  */
-export const getUpdatedSelection = (newTokenLabels, selectedItems) =>
-  selectedItems.filter((item) => newTokenLabels.includes(itemToString(item)));
+const defaultSummaryFormatter = ({
+  selectedItems,
+  getSelectedItemProps,
+  removeSelectedItem,
+  disabled,
+  label,
+}) => {
+  if (selectedItems.length === 0) {
+    return label;
+  }
+  return (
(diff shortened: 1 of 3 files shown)


src/MultiSelect/index.js
```diff
@@ -97,6 +113,14 @@ const MultiSelect = ({
     },
   });
 
+  // Only sync internal state if the component is controlled.
```
JohnathanWeisner, line 116: There was a bug prior where if the state outside the component updated then the MultiSelect component wouldn't change its selections. This fixes that.

src/MultiSelect/index.js
```diff
@@ -13,7 +14,7 @@ import { createUseLayerContainer } from "src/util/dom";
 const noop = () => {};
 
 /**
- * @param item JXS node
+ * @param item JSX node
```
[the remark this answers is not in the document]
xiexiewenting: towards https://github.com/narmi/banking/issues/13151

src/MultiSelect/index.js
davidneto92: Are there many/any cases where we would use this component as uncontrolled, or outside of a form?

JohnathanWeisner: All prior uses were technically uncontrolled with initial values being set. I'm not sure if that was intended though.

src/MultiSelect/index.js
```diff
@@ -178,6 +201,73 @@ const MultiSelect = ({
 
   const hasSelectedItems = selectedItems.length > 0;
 
+  // Compute summary text for "summary" kind.
+  let summaryText = "";
```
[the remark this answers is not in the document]
JohnathanWeisner: I went with individual useMemos for each since the dependencies were very diff.

src/MultiSelect/index.scss
```diff
@@ -46,3 +46,10 @@
   gap: var(--space-xs) var(--space-s);
   margin-right: var(--space-s); // make room for the right caret
 }
+
+.nds-multiselect-clearAll {
+  position: absolute;
+  right: 38px;
+  top: 50%;
+  transform: translateY(-50%);
+}
```
akdetrick, line 55: (optional)

We may want to add some extra room on the left for this absolutely positioned element when the clear button is rendered
Screenshot 2025-03-27 at 10 37 33 AM

akdetrick (commented):
## Optional suggestion

What if instead of having a `kind` prop to flag when to use the formatter, this component _always_ used the `summaryFormatter`?

The default value of `summaryFormatter` could be a function that maps the strings to `FieldToken` components to cover existing behavior. The signature of the formatter may need to change so it has access to label values and ids for dismissal, which would make it more flexible for custom use later on.

Users would then be able to override that function with whatever they want, exactly as you have it in the story.

src/MultiSelect/index.scss
JohnathanWeisner: 👍 Good call.

akdetrick (approved):

PR #1582, opened 2025-04-04
fix(Combobox): persist selection when users select an action

akdetrick:
Closes NDS-1243

The [recent fix we merged for this issue](https://github.com/narmi/design_system/pull/1581) didn't completely work. I found that the enhancements we have added over time sort of painted us into a corner with complexity around managing downshift events.

This PR shifts most of the logic we need to accomplish the desired behaviors (unit tests pass) to the `stateReducer` we pass into the `useCombobox` config. 

This not only makes the code easier to work with, but resolves some of the chicken/egg problems that were inherent with the approach of using event hooks in addition to the state reducer to manage some of the same behaviors (e.g. blur).

### Code changes
- simplify and change order of things in downshift's `onInputValueChange` event hook
- move all other logic about combobox state to the state reducer

### How to review this
1. Help me make the code better if there are any opportunities (_side by side diff will be easier to check_)
2. Open the storybook preview that gets commented on this PR. Really kick the tires. Does this act like what you think a combobox acts like? Any frustrations or awkwardness in interaction?
3. Test the `WithActions` story of `Combobox`. This is a perfect test case for the original bug we aimed to solve - if the user has selected an item, that item should stay selected after clicking an action item.

### DIFF
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -203,18 +203,8 @@ const Combobox = ({
         return;
       }
 
-      const itemMatchingInput = items.find(
-        (item) =>
-          inputValue === item.props.searchValue ||
-          inputValue === item.props.value,
-      );
-
-      // once a selection is made, show all options in the dropdown
-      if (itemMatchingInput) {
-        setDisplayedItems(items);
-
-        // filter by input value if filtering is enabled
-      } else if (!disableFiltering) {
+      // Filtering based on inputValue
+      if (!disableFiltering) {
         const actionItems = items.filter(isAction);
         const filteredItems = filterItemsByInput(
           items.filter((item) => !isAction(item) && isSelectable(item)),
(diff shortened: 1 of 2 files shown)


src/Combobox/index.js
```diff
@@ -150,7 +150,6 @@ const Combobox = ({
   testId,
   renderEndContent = defaultRenderEndContent,
 }) => {
-  const inputRef = useRef(null);
```
akdetrick, line 153: we no longer need this to manually blur the input

src/Combobox/index.js
```diff
@@ -175,7 +174,9 @@ const Combobox = ({
   }
 
   const [displayedItems, setDisplayedItems] = useState(items);
-  const [clearOnNextInput, setClearOnNextInput] = useState(false);
```
akdetrick, line 178: no longer needed - input clearing is all handled in the `stateReducer` and `onInputValueChange`

akdetrick:
It may have been ill advised to reset the filtered list when an item is selected because you can't see the checkmark when you reopen the dropdown.

According to WCAG, it [should only show the filtered item](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/)

josh-narmi (approved):
Tested this manually, and it all seemed to work correctly!

PR #1583, opened 2025-04-04
feat(MenuButton): add positioning api from Popover

akdetrick:
Closes NDS-1278

Adds popup positioning control to `MenuButton` using the same props API as `Popover`.
Needed for `SplitButton` component in review.

See `MenuButton` story in the storybook preview build.

### DIFF
--- a/src/MenuButton/index.js
+++ b/src/MenuButton/index.js
@@ -31,6 +31,9 @@ const MenuButton = ({
   renderTrigger,
   triggerIcon = "more-vertical",
   showDropdownIndicator = false,
+  side = "bottom",
+  alignment = "start",
+  offset = 2,
   children,
 }) => {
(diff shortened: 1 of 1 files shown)


src/MenuButton/index.js
```diff
@@ -41,10 +44,9 @@ const MenuButton = ({
     onOutsideClick: () => {
       setIsOpen(false);
     },
-    preferX: "right",
-    preferY: "bottom",
```
XihaoLuo, line 45: I'm reading the docs in https://github.com/everweij/react-laag?tab=readme-ov-file#uselayer and it seems like these only have an effect when `auto: true`? if so, ok to remove.

XihaoLuo (approved):
Makes sense.

josh-narmi (approved):

PR #1593, opened 2025-04-16
fix(Combobox,Select): force a rerender when children change

akdetrick:
Closes NDS-1350

- memoize children to array processing in `Combobox` and `Select`
- ensure component rerenders the items in the dropdown when `children` change in both components
- add stories as test cases

Apr-16-2025 17-39-08

### DIFF
--- a/src/Combobox/index.js
+++ b/src/Combobox/index.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useMemo, useEffect } from "react";
 import PropTypes from "prop-types";
 import cc from "classcat";
 import iconSelection from "src/icons/selection.json";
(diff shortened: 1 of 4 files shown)


src/Combobox/index.js
```diff
@@ -160,6 +163,8 @@ const Combobox = ({
       ? []
       : allChildren.filter(({ props }) => "value" in props || "text" in props);
 
+  console.info(items);
```
akdetrick, line 166:
```suggestion
```

lol

src/Select/index.stories.js
```diff
@@ -294,6 +294,10 @@ InAForm.parameters = {
 
 export const Controlled = () => {
   const [value, setValue] = useState(null);
+  const [items, setItems] = useState([
+    { value: "checking1234", label: "Checking (1234)" },
+    { value: "savings4321", label: "Savings (4321)" },
```
akdetrick, line 299: I updated the controlled examples in both components to use a state var instead - this is much closer to real world usage

blooies (approved):
lgtm! thanks for adding!!

PR #1596, opened 2025-04-17
fix(Drawer): enable focus trap; restore focus rings on mobile controls

akdetrick:
Closes NDS-894

Like `Dialog`, we need a focus trap so that focus moves circularly through all focusable elements in Drawer when Drawer is open (including drawer controls).

#### Mobile control buttons changes
I noticed on mobile, the controls don't have a focus ring. This is because we were absolutely positioning the button _contents_ (the spans rendering the icons) rather than the button itself. Absolute positioning takes them out of document flow, causing the actual button elements that are the focusable to have no natural width or height, hiding the focus ring from view.

Fixed by moving positioning rules to the buttons themselves rather than the icons.

##### Note the trigger button behind the shim is not included in the focusable items
Apr-17-2025 17-02-02
Apr-17-2025 17-02-32

### DIFF
--- a/src/Drawer/index.js
+++ b/src/Drawer/index.js
@@ -6,6 +6,7 @@ import cc from "classcat";
 import useBreakpoints from "../hooks/useBreakpoints";
 import useMountTransition from "./useMountTransition";
 import useLockBodyScroll from "../hooks/useLockBodyScroll";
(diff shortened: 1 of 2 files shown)


josh-narmi (approved):
Excellent!

PR #1597, opened 2025-04-18
fix(Button): style fixes for 'plain' kind

akdetrick:
Closes NDS-1353

When we merged `SplitButton` in `4.18.0`, the refactoring of button styles in to Sass placeholders missed two important things with the `plain` variant of `Button`...

- I forgot to define the variant class in `Button/index.scss`
- I forgot that `kind="plain"` buttons need an override for font weight instead of inheriting from `sizes` placeholders

This PR addresses both to restore `Button kind="plain"` to it original styling.

### Verifying the fix
You can check this older, pre-regression storybook deployment for the **expected** styles: https://60620d422ffdf100216415b2-vnjujhnpnl.chromatic.com/?path=/docs/components-button--docs

For the **actual** styles in this PR, see the Storybook comment in this conversation thread.

They should match.

### DIFF
--- a/src/Button/index.scss
+++ b/src/Button/index.scss
@@ -94,6 +94,10 @@
   @extend %btn-state-loading;
 }
 
+.nds-button--plain {
+  @extend %btn-kind-plain;
+}
+
 /**
 ️* ⚠️ DEPRECATED ⚠️
 * Will be removed in a future release.
--- a/src/Button/index.scss
+++ b/src/Button/index.scss
@@ -5,9 +5,9 @@
 
 .nds-button,
 .nds-button.resetButton {
+  cursor: pointer;
   display: inline-flex;
   position: relative;
-  cursor: pointer;
   line-height: var(--font-lineHeight-bigText);
   text-align: center;
   justify-content: center;
@@ -38,12 +38,6 @@
 }
 
 /* Kind variants */
-.nds-button--primary,
-.nds-button--secondary,
-.nds-button--tonal {
-  @extend %btn-kind-base;
-}
-
 .nds-button--primary,
 .nds-button--primary.resetButton {
   @extend %btn-kind-primary;
--- a/src/Button/sizes.scss
+++ b/src/Button/sizes.scss
@@ -3,6 +3,10 @@
   font-weight: var(--font-weight-semibold);
   min-height: rem(24px);
   border-radius: rem(12px);
+  .nds-button-content {
+    margin: rem(3px) rem(16px);
+    max-width: rem(200px);
+  }
 }
 
 %btn-size-s {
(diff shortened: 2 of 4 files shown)


src/Button/variants.scss
```diff
@@ -69,13 +69,8 @@
   }
 }
 
-/**
-️* ⚠️ DEPRECATED ⚠️
-* Will be removed in a future release.
-*/
```
akdetrick, line 75: Not actually deprecated! This message was only intended for `kind="menu"`

src/Button/index.scss
```diff
@@ -38,12 +38,6 @@
 }
 
 /* Kind variants */
-.nds-button--primary,
-.nds-button--secondary,
-.nds-button--tonal {
-  @extend %btn-kind-base;
-}
```
akdetrick, line 45: No longer needed - the `%btn-kind-base` placeholder was removed in this PR because those styles are coupled to the `size` prop.

src/Button/index.scss
```diff
@@ -94,6 +88,10 @@
   @extend %btn-state-loading;
 }
 
+.nds-button--plain {
+  @extend %btn-kind-plain;
+}
+
```
akdetrick, line 94: This is the main fix. 🙃

src/Button/sizes.scss
```diff
@@ -3,18 +3,30 @@
   font-weight: var(--font-weight-semibold);
   min-height: rem(24px);
   border-radius: rem(12px);
+  .nds-button-content {
+    margin: rem(3px) rem(16px);
+    max-width: rem(200px);
+  }
```
akdetrick, line 9: `.nds-button-content` exists to set the constraints of the label area of the button. The values here are related to `size`, so they have been moved to the appropriate size placeholders here.

src/Button/variants.scss
```diff
@@ -61,25 +41,16 @@
   }
 }
 
-%btn-kind-negative {
-  @extend %_btn-kind-plain-base;
-  color: var(--theme-primary);
-  &:hover {
-    color: var(--theme-primary);
-  }
-}
```
akdetrick, line 70: Moved this _below_ the `%_btn-kind-plain-base` placeholder for clarity.

src/Button/variants.scss
```diff
@@ -110,6 +86,7 @@
   text-align: left;
   color: var(--font-color-primary);
   display: block;
+  font-weight: var(--font-weight-semibold);
```
akdetrick, line 89:
```suggestion
```

src/SplitButton/index.scss
```diff
@@ -51,12 +51,6 @@
   width: rem(40px);
 }
 
-.nds-splitButton-action--primary,
-.nds-splitButton-action--secondary,
-.nds-splitButton-action--tonal {
-  @extend %btn-kind-base;
-}
-
```
akdetrick, line 59: This placeholder is no longer necessary, as the `sizes` placeholders now contain the necessary styles that were being applied here.

graphicsforge (approved):
Hard to QA how it'll look in banking without this merged, so let's merge this first! 👍

PR #1580, opened 2025-04-01
feat(SplitButton): add SplitButton component

akdetrick:
Closes NDS-981

### `Button`
- Move state, variant, and size styles for button into [sass placeholders](https://sass-lang.com/documentation/style-rules/placeholder-selectors/) to make them reusable
- Update styles for `nds-button` to use placeholders

### `SplitButton`
- Add component
- Consume `Button` placeholder selectors to style custom trigger to match the main button

## What are Sass placeholders?

Placeholders allow us to share the same style rules we apply to `Button` sizes, variants, and states with `SplitButton`.

Any selector in Sass that starts with `%` is a "placeholder". The rule block defined under a placeholder will **not** be emitted as CSS. Placeholders allow us to share blocks of CSS via the `@extend` directive. 

What's actually happening under the hood is that Sass hoists the selector being extended to a multiple selector that contains the placeholder rule block. 

**scss**
```scss
%warning {
   color: lightpink;
   background: red;
}

.formError {
   @extend %warning;
   font-size: 1rem;
}

.errorSummaryHeading {
   @extend %warning;
   font-size: 3rem;
}
```

**compiles to this CSS**
```css
.errorSummaryHeading,
.formError {
   color: lightpink;
   background: red;
}

.formError {
   font-size: 1rem;
}

.errorSummaryHeading {
   font-size: 3rem;
}
```

## Props interface

`SplitButton` follows our pattern of subcomponents. Here's a good basic usage example:

```jsx
  <SplitButton label="Send" onClick={handleMainButtonClick}>
    <SplitButton.Menu>
      <SplitButton.MenuItem
        label="Schedule"
        onSelect={handleItemSelect}
      />
      <SplitButton.MenuItem
        label="Save as draft"
        onSelect={handleItemSelect}
      />
    </SplitButton.Menu>
  </SplitButton>
  ```
  
My thinking here is that this keeps us open to new behaviors. Should our design team design a split button where the right hand action makes a [unicorn](https://www.cornify.com/) fly across the screen, it would be implemented as a `<SplitButton.Unicorn>` subcomponent.

The root `SplitButton` component manages its children. While a user could technically pass multiple subcomponents, `SplitButton` will only render the first subcomponent it encounters.

Screenshot 2025-04-15 at 1 09 55 PM

PR #1599, opened 2025-04-22
fix(SplitButton): export SplitButton

elliot-wilson:
Follow-up to https://github.com/narmi/design_system/pull/1580

Makes `SplitButton` available for export.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -38,6 +38,7 @@ import Sidebar from "./Sidebar";
 import Slider from "./Slider";
 import Snackbar from "./Snackbar";
 import Spinner from "./Spinner";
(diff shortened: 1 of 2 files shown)


src/SplitButton/index.tsx
```diff
@@ -98,7 +98,6 @@ const SplitButton = ({
           <Popover
             // eslint-disable-next-line jsx-a11y/no-autofocus
             autoFocus
-            // @ts-expect-error Popover is not TS yet
```
elliot-wilson, line 101: This line was causing a TS error (unused warning)

XihaoLuo (approved):

PR #1600, opened 2025-04-23
fix(SplitButton): don't require props that have a default

elliot-wilson:
I noticed in NDS 4.18.4 that I need to supply `kind="primary"` and `size="m"` for SplitButton, despite these props having defaults.

If that's intentional, feel free to close this PR.

As a drive-by, I also cleaned up two typos in these files.


akdetrick (approved):
Excellent find, thanks!

PR #1603, opened 2025-04-24
fix(Button): flow plain variant as text

akdetrick:
Closes NDS-1400

Fixes text alignment of `plain` variant of `Button`.

This PR also adds a new story that will be a better test against future regressions:

Screenshot 2025-04-24 at 2 29 21 PM
Screenshot 2025-04-24 at 3 17 48 PM


src/Button/variants.scss
```diff
@@ -89,7 +99,7 @@
   font-size: var(--font-size-l) !important;
 
   .nds-button-content {
-    margin: rem(10px) 0;
```
akdetrick, line 92: Peter and I had the same fix - need to update from main and this diff line will go away

blooies (approved):
ty!!

Issue #1605, opened 2025-04-29
[Combobox]: Empty input value will result in max depth exception


agreatscott:
### Component/Module

Combobox

### Reproduction Steps

1. Set up a Combobox with categories (storybook story exists already)
2. Enter a value
3. Clear that value
4. See max depth exception in the console

### Expected Result

No max depth exception

### Actual Result

max depth exception

PR #1606, opened 2025-04-29
fix(combobox): resolve infinite re-render issues with combobox

agreatscott:
for https://github.com/narmi/design_system/issues/1605

Notably, I don't see this in prod builds - it's not reproducible https://narmi.github.io/design_system/?path=/story/components-combobox--with-categories, but it is reproducible running the same story locally. 

When clearing an input, we see an error in the console detailing that the max update depth has exceeded. 

---

On further testing, there were multiple cases where we would introduce the infinite update exception, mostly oriented around how we did not make the `stateReducer` a pure function. This PR removes the blur handling and selected item change handling out of that function and into other functions provided by downshift

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@narmi/design_system",
-  "version": "4.18.10",
+  "version": "4.19.0-beta.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@narmi/design_system",
-      "version": "4.18.10",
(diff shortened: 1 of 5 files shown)


akdetrick (commented):
Excellent find - I think we may be running into yet another bug that we can only observe in certain builds, where state bundling is not guaranteed to happen the exact same way between prod and dev.

Moving things out of the state reducer avoids this problem by making sure we're not modifying state from multiple touch points.

akdetrick (approved):
Fully controlled example preserves the blur behavior - looks good to go

PR #1609, opened 2025-04-30
fix(Button): allow overflow in plain and negative buttons

akdetrick:
Relates to NDS-1400

https://linear.app/narmi/issue/NDS-1400/[design-system]-plain-button-variant-minor-regression#comment-c13abbb3

Prevents base `overflow: hidden` rule on buttons from cutting off long text in plain and negative buttons.


xiexiewenting (approved):
looks great!

### before
Screenshot 2025-04-30 at 6 14 32 PM

### after
Screenshot 2025-04-30 at 6 14 42 PM

PR #1615, opened 2025-05-05
fix(MenuButton): allow menu toggling

josh-narmi:
- [NDS-1366: SplitButton with MenuButton does not toggle](https://linear.app/narmi/issue/NDS-1366/splitbutton-with-menubutton-does-not-toggle)

### DIFF
--- a/src/MenuButton/index.js
+++ b/src/MenuButton/index.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from "react";
+import React, { useState, useEffect, useRef, useCallback } from "react";
 import PropTypes from "prop-types";
 import cc from "classcat";
 import { useLayer } from "react-laag";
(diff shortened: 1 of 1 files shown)


src/MenuButton/index.js
```diff
@@ -72,9 +77,17 @@ const MenuButton = ({
...
   };
 
+  const pressButton = useCallback(() => {
+    if (canToggleCloseRef.current) {
+      closeMenu();
+    } else {
+      canToggleCloseRef.current = true;
+    }
+  }, [closeMenu]);
+
```
josh-narmi, line 90: Kind of silly, but since `onClick` is triggered on first menu open, and `isOpen` is already set to true by then, we need a second, non-stateful variable to keep track of toggle-ability.

elliot-wilson (approved):
weird, but it works!

PR #1618, opened 2025-05-07
fix(AutocompleteModal): rerender list when children change

akdetrick:
closes NDS-1339

Ensure AutocompleteModal responds to changes in `children` (programmatically adding new items).

- Update "With Action" story of AutocompleteModal to cover this behavior
- Memoize children in the root `AutocompleteModal`
- Add a `useEffect` to ensure the internal `Autocomplte` rerenders when `items` passed in changes

May-07-2025 14-04-55


src/AutocompleteModal/AutoComplete.tsx
```diff
@@ -45,6 +45,12 @@ const AutoComplete = ({
   footerContent,
 }: AutoCompleteProps) => {
   const [items, setItems] = useState(allItems);
+
+  // update if items change programmatically
+  useEffect(() => {
+    setItems(allItems);
+  }, [setItems, allItems]);
```
josh-narmi, line 52: nit: I don't think you need to pass the standard `useState()` setter to the dependency array

src/AutocompleteModal/index.tsx
```diff
@@ -39,7 +39,7 @@ const AutocompleteModal = ({
   const [selectedItem, setSelectedItem] = useState<React.ReactElement | null>(
     null,
   );
-  const items = React.Children.toArray(children);
+  const items = useMemo(() => React.Children.toArray(children), [children]);
```
josh-narmi, line 42: Love to memo!

josh-narmi (approved):

PR #1622, opened 2025-05-08
feat(colors): add amethyst, aka AI purple

elliot-wilson:
Adds the "AI purple" used in various new staff designs to mark AI features:

image

The yellowish color in the gradient here is the existing `var(--color-sand)`.

image


akdetrick (approved):
💜

PR #1631, opened 2025-05-12
fix(radioButtons): allow passing inputs to RadioButtons children

davidneto92:
Our current implementation of the `<RadioButtons />` component allows for adding children for a radio option through options > details. However, if the children contain `<input />` tags the styling for the RadioButtons makes the children inoperable.

This PR updates the CSS to allow full interaction with children components, and adds a `allowChildrenInput` prop to skip the `handleChange` function from the parent div when instructed.

Recording of updated behavior showing children input interactions, and unaffected behavior of other styles.

https://github.com/user-attachments/assets/9436530f-6996-468b-a6d1-24e7963930e8


akdetrick (changes requested):
Is this something we can solve with composition of `Radio` instead? It takes arbitrary JSX as children:
https://narmi.github.io/design_system/?path=/docs/components-radio--docs

I'm slightly concerned with rendering two inputs within one `label` element, so we may want to refactor `Radio` to render the `type="radio"` input outside of the label.

davidneto92:
Not needed per discussion. Thanks!

PR #1632, opened 2025-05-12
fix(Button): consistently align 'menu' variant

akdetrick:
closes NDS-1503

The long deprecated `menu` variant of `Button` is still being used in the OLB header, and some CSS in banking is also targeting these elements. This is a bad situation, so we will remove the `menu` kind in the next major release.

The changes here ensure menu buttons are properly aligned regardless of what element they render via `as`, or what other banking styles are targeting these elements.

#### With our overrides
Screenshot 2025-05-12 at 3 27 32 PM

#### Current
Screenshot 2025-05-12 at 3 27 39 PM

### DIFF
--- a/src/Button/variants.scss
+++ b/src/Button/variants.scss
@@ -102,9 +102,16 @@
 %btn-kind-menu {
   text-align: left;
   color: var(--font-color-primary);
-  display: block;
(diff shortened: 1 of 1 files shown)


josh-narmi (approved):
Can't wait to remove these deprecated elements!

PR #1633, opened 2025-05-12
build(release): add deprecation management

akdetrick:
closes NDS-1363

In draft while testing...


release.config.js
```diff
@@ -4,15 +4,26 @@
...
       "@semantic-release/commit-analyzer",
       {
         preset: "conventionalcommits",
       },
     ],
+    [
+      "semantic-release-npm-deprecate-old-versions",
+      {
+        rules: [
+          checkVersion, // our own custom rule, so we can manage deprecated version ranges
```
akdetrick, line 22: according to docs, this _should_ be fine, but....

```
AggregateError: 
    TypeError: Cannot read properties of undefined (reading 'rule')
        at /home/runner/work/design_system/design_system/node_modules/semantic-release-npm-deprecate-old-versions/dist/configuration-loader.js:23:28
        ```
investigating the cause...

akdetrick:
Not using this approach. Closing.

PR #1635, opened 2025-05-15
chore: export beta `Table` component

akdetrick:
Closes NDS-1521

Adds a v0 `Table` component to NDS. I say "v0" because Storybook will omit documentation for this until we feel it's ready.

In the meantime, the ACH activity project may important use this component for a test drive as we make improvements before fully releasing the `Table` component publicly with documentation.

## Props interface
Columns will never exceed 12, and there could be any number of rows. The main lever of control for table layout is in defining the width and visibility of columns over different viewport breakpoints.

- `colVisiblity` - an array of strings, where each string declares which breakpoint must be satisfied to show the col
- `colLayout` - and object of breakpoint name keys to `grid-template-column` values

Combined, this allows us to control which columns are visible at any breakpoint _and_ change the column layout accordingly.

### Why `grid-tempalte-columns`
CSS has already designed an interface for configuring columns, and it's this property. It gives us a lot of control over column widths. All of the following are valid for setting column widths:

- `minmax(100px, 300px)`_what it says on the tin_
- `40ch` _hard width based on characater count of 40 for the given font size and weight_
- `minmax(1fr, 60ch)` _minimum 100% fraction of remaining space; maximum of 60 chars_
- `50% 50%` _percentages if you're old school_
- `1f 1fr` _an `fr` is a fraction of the remaining space - here it's equivalent to `50% 50%`_
- `2fr 1fr` _all fractions; first col takes 2/3 width, third takes remaining_
- `repeat(12, 1fr)` _if all 12 of your columns are equal width_
- `min-conent` _size column to minimum content (useful for checkboxes, etc)_
- `max-content` _column becomes as wide as the widest entry in the column_
- `137px`, `2rem` _any value is valid if you want to get weird with it_

**Putting it together:*** Three equal columns and a kebab menu on the end (4 total)
`grid-template-columns: repeat(3, 1fr) min-content;`

## Preview
https://github.com/user-attachments/assets/12e942a3-f9ce-4942-8ff7-95a6849fa546

## How to review
The storybook preview **will not** contain the `Table` stories. You must run this branch locally, and rename `src/Table/index.stories-hidden.js` to `src/Table/index.stories.js` before running `npm run dev`.

## What is a `subgrid`?

Much like Flexbox, adding a `display: grid` to an element changes its _layout mode_, making its direct children implicit grid children. The problem is, you often want some elements to contribute to the calculation of those grid tracks when the element may not be a direct descendant.

Until `subgrid`, `display: contents` was our only option. It's problematic for reasons I won't get into here. As of right now, `subgrid` is widely supported and is used as a value keyword to inherit grid properties and share grid tracks between elements.

### What does it mean to "participate" in grid tracks?

Consider this property for a 3 column grid: `grid-template-columns: 1fr max-content 1fr`

The `max-content` value means that the middle column will be as wide as the widest content contained in that column. Any element that shares this grid track is an active participant, as it is both influenced by the width of other cells in its column track, but can also influence it.

**tl;dr** it took 25 years to make CSS do what an HTML `table` does (to be fair, with much more flexibility now)


src/Table/Cell.tsx
```diff
@@ -0,0 +1,34 @@
...
+  _colIndex?: number;
+}
+
+const Cell = ({ children, textAlign = "start", _colIndex = 0 }: CellProps) => {
+  const { section } = useContext(TableSectionContext);
+  const { currentBreakpoint, colVisibility } = useContext(ColVisibilityContext);
+
+  const role = section === "header" ? "columnheader" : "cell";
+  const minBreakpoint = colVisibility[_colIndex];
+  const isVisible = isBreakpointSatisfied(minBreakpoint, currentBreakpoint);
```
akdetrick, line 25: Example: 
The user has passed in a `"*"` for this column, indicating it should be hidden. This value, the `minBreakpoint`, is compared against the `currentBreakpoint`. In this case, it would return true so the column is always visible.

src/Table/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+    >
+      <div
+        role="table"
+        className={cc(["nds-table", `nds-table--${rowDensity}`])}
+        /**
+         * We apply a CSS value for grid-tempalte-columns on the root
+         * of this table component, so the grid column tracks may be
+         * shared among all rows via `subgrid`.
+         */
+        style={{ gridTemplateColumns }}
```
akdetrick, line 123: This is where the magic happens. Through some layers of validation above, we determine a CSS value for `grid-template-columns` and set it on the table root element. 

Defining the table columns on root allows cells from different table sections (header, body) to participate in and share the same grid tracks via `subgrid`.

src/Table/util/grid.ts
```diff
@@ -0,0 +1,17 @@
+import type { ColLayoutConfig, CSSValue, ViewportBreakpoint } from "..";
+
+/**
+ * Determine the final `grid-template-columns` value to apply
+ * to a rowgroup element shit.
```
akdetrick, line 5:
```suggestion
 * to a rowgroup element.
```

?

src/Table/index.scss
```diff
@@ -0,0 +1,71 @@
+.nds-table {
+  display: grid;
+  grid-auto-rows: auto;
+  row-gap: 0;
+}
```
akdetrick, line 5: Column configuration is applied via inline style on `nds-table`.

src/Table/index.scss
```diff
@@ -0,0 +1,71 @@
...
+
+.nds-table-body,
+.nds-table-header {
+  display: grid;
+  grid-auto-rows: subgrid;
+  grid-template-columns: subgrid;
+  grid-column: 1 / -1; // participate in all parent columns
```
akdetrick, line 12: In CSS grid properties, declaring grid tracks works similarly to python list comprehension.

This is saying "I want the body and header grids to span from the first to the last column of the parent grid, regardless of how many there are". `-1` == last

src/Table/index.stories-hidden.js
```diff
@@ -0,0 +1,225 @@
...
+import MenuButton from "../MenuButton";
+import Checkbox from "../Checkbox";
+
+export const Overview = () => (
+  <div className="padding--all bgColor--blueGrey">
+    <div className="margin--bottom--xl">
+      <Table
+        colVisibility={["*", "m", "*", "l"]}
+        colLayout={{
+          s: "invalid lol", // falls back to even 1fr cols
```
akdetrick, line 20: invalid CSS values passed here will fall back to something sensible and trigger a `console.warn`.

src/Table/util/breakpoint.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const breakpointOrder = ["s", "m", "l"] as const;
+type OrderedBreakpoint = (typeof breakpointOrder)[number];
+
+/**
+ * We use "mobile-first" breakpoints, meaning that
+ * "m" is `true` for medium viewports and larger.
+ * (e.g. "l" satisfies "m")
+ */
+export const isBreakpointSatisfied = (
```
akdetrick, line 11: I gave this some thought, and I feel that we should move this to the `useBreakpoints` hook.

We'd add two items to the returned object in `useBreakpoints`:

1. `isBreakpointSatisfied(breakpointName)` - compare a bp name to current
2. `currentBreakpoint` - name of current or closest breakpiont

src/Table/Cell.tsx
blooies: i thought `*` means it should be visible for all browser sizes?

src/Table/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+      value={{
+        colVisibility,
+        currentBreakpoint: currentBreakpoint as ViewportBreakpoint,
+      }}
+    >
+      <div
+        role="table"
+        className={cc(["nds-table", `nds-table--${rowDensity}`])}
+        /**
+         * We apply a CSS value for grid-tempalte-columns on the root
```
blooies, line 119: small typo `grid-template-columns`

blooies (approved):
this looks so good!! so exciting! thanks for adding all the customizations and the descriptive comments!

PR #1648, opened 2025-05-31
chore: update chromatic workflow

akdetrick:
Relates to NDS-1401

- Cleans up where chromatic configs/flags are set are passed
- Set options on the chromatic job to ensure it fails [if the exit code is not `0`](https://www.chromatic.com/docs/cli/#exit-codes)
- Separates the commenting step into its own job, so we can fail the chromatic job while still executing the next job


akdetrick:
Check not running. CLosing.

PR #1660, opened 2025-06-02
fix(Row): rm outdated item hight fix for IE

akdetrick:
Closes NDS-1538

[As Elliot discovered](https://linear.app/narmi/issue/NDS-1538/regression-row-alignment-has-broken-potentially-across-multiple#comment-0a1020e2), there's a case where a Row containing items with certain height rules may have incorrect alignment due to the `height: 100%` rule on row items.

This rule was added by me, based on old habits for browsers we no longer need to support.

We should not expect or accept _any_ changes in snapshot tests.


elliot-wilson (approved):
💪

PR #1661, opened 2025-06-03
Add `Chip` and `Count` components

akdetrick:
### `Chip`

The `Tag` component is based on some older, more ridged design standards. This new component reflects the tag-like Chip that designers are now using in Figma.

#### Interaction
A `Chip` can either be clickable in its entirety as a `button` **OR** it can contain a close button within it. Never both. The component overrides the close button when `onClick` is passed to make the entire area clickable.

#### Variants
There are no size variants decided by the design team yet. Style variants are color-only.
Screenshot 2025-06-05 at 2 04 36 PM

### `Count`
It's just a rounded div with text and color variants, but following the new Figma standard for displaying status counts throughout our products.


src/Chip/index.scss
```diff
@@ -0,0 +1,71 @@
...
+  justify-content: center;
+}
+
+.nds-chip {
+  @extend %chipAlignedContent;
+  border-radius: rem(24px);
+  font-size: var(--font-size-s);
+  padding: rem(2px) var(--space-s);
+  width: fit-content;
+  max-width: 60ch; // label truncates when max constraint is reached.
```
akdetrick, line 14: It's 60 chars width for the entire chip - so more like 40-50 for the label, depending on icons and such

src/Chip/index.scss
```diff
@@ -0,0 +1,71 @@
...
+    font-weight: var(--font-weight-semibold);
+  }
+  &:focus,
+  &:hover {
+    transform: scale(1.2);
+  }
+  &:active {
+    transform: scale(1);
+  }
+}
```
akdetrick, line 71: This is a Loupe effect on hover and click. In absence of a defined hover style, I went with this to keep the chip from getting cluttered. Design has approved, but this may still be subject to tweaks and changes.

kallmeehkev (approved):
looks good!

PR #1663, opened 2025-06-04
chore(Table): update beta component with HeaderCell

akdetrick:
Closes NDS-1571

Adds `Table.HeaderCell` subcomponent.

⚠️ This won't be available in the storybook preview as we're hiding the docs until the component is stable.

This commit will not trigger a release because it is type `chore`. This change will be included in the next release this sprint, but will not be included in release notes. This is intentional - we only want to publish details on this component once it's ready and we can make a `feat(Table)` commit.

---

The focus ring is present because this specific cell has an `onClick` handler passed, which renders a `button` instead of a plain `div`. This allows users to access HeaderCells by keyboard, as I've done in the screenshot below:

Screenshot 2025-06-04 at 5 40 55 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -134,10 +134,11 @@
       }
     },
     "node_modules/@babel/cli": {
-      "version": "7.26.4",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.26.4.tgz",
-      "integrity": "sha512-+mORf3ezU3p3qr+82WvJSnQNE1GAYeoCfEv4fik6B5/2cvKZ75AX8oawWQdXtM9MmndooQj15Jr9kelRFWsuRw==",
+      "version": "7.27.2",
+      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.27.2.tgz",
+      "integrity": "sha512-cfd7DnGlhH6OIyuPSSj3vcfIdnbXukhAyKY8NaZrFadC7pXyL9mOL5WgjcptiEJLi5k3j8aYvLIVCzezrWTaiA==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "@jridgewell/trace-mapping": "^0.3.25",
         "commander": "^6.2.0",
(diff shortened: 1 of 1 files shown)


src/Table/Body.tsx
```diff
@@ -1,12 +1,9 @@
 import React from "react";
-import TableSectionContext from "./util/tableSectionContext";
 
 const Body = ({ children }: { children: React.ReactNode }) => {
   return (
     <div className="nds-table-body" role="rowgroup">
-      <TableSectionContext.Provider value={{ section: "body" }}>
```
akdetrick, line 7: This context was used for one thing only - telling `Cell` where in the table tree it lives by section (either "body" or "header").

Now that we no longer share `Cell` between the body and header of the table, we no longer need this context for conditional aria roles etc

src/Table/HeaderCell.tsx
```diff
@@ -0,0 +1,64 @@
...
+}: HeaderCellProps) => {
+  const { currentBreakpoint, colVisibility } = useContext(ColVisibilityContext);
+  const minBreakpoint = colVisibility[_colIndex];
+  const isVisible = isBreakpointSatisfied(minBreakpoint, currentBreakpoint);
+  const isButton = typeof onClick === "function";
+
+  if (!isVisible) return null;
+
+  return (
+    <div aria-live="polite">
```
akdetrick, line 41: I'm thinking this is how it will work:

1. user clicks `HeaderCell` and our `onClick` handler runs
2. The consuming component of `Table` updates and rerenders with the sort order changed for the affected column
3. The `HeaderCell` has new copy and/or sort indicator. If we include an `aria-label` on the sort indicator icon, changes in its value will be announced

tl;dr - click sort action header cell, hear that the sort changed if you're using a screen reader

akdetrick: I think eventually we want to include standard props for rendering sorting and filtering indicators, but for now it may be best to compose it in manually before we make too many assumptions

blooies (approved):
looks good, ty for adding!

PR #1675, opened 2025-06-16
demonstrate hover behavior for MenuButtons wrapped in Tooltips

elliot-wilson:
https://github.com/user-attachments/assets/514787a1-5a34-4c01-87b1-f6f1fa9de300

### DIFF
--- a/src/Tooltip/index.stories.js
+++ b/src/Tooltip/index.stories.js
@@ -2,6 +2,7 @@ import React, { useState } from "react";
 import Tooltip from "./";
 import Button from "../Button";
 import TextInput from "../TextInput";
+import MenuButton from "../MenuButton";
 
 const Template = (args) => (
   <div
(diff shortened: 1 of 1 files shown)


akdetrick:
@elliot-wilson 

Thanks for putting this draft up to create a case for the issue, this really is the best way to debug NDS problems.

I added one more suggestion - it's a little silly, but doesn't really go against any of the intended usage of controlled Tooltips or the `renderTrigger` prop. 

The drawback is that we'd have to replicate the MenuButton trigger inside the tooltip children

akdetrick:
### Next steps
- [ ] remove debugging stories except for accepted solution (assuming it works)
- [ ] rename story to `ToolTipAsTrigger`
- [ ] move that story to `MenuButton` stories

Then we can at least have the pattern documented and checked by visual tests.

PR #1681, opened 2025-06-16
Feat/current breakpoint tables

akdetrick:
Closes NDS-1536
Closes NDS-1598

Updates our breakpoints hook to meet the requirements of `Table`.

- Add a new return value for `useBreakpoints` that returns the name of the largest satisfied breakpoint. Can be used to determine the "current" breakpoint name
- Update `Table` component usage of `useBreakpoints` to use this new standard feature and remove ad-hoc util

### A note on "mobile-first" breakpoints
NDS `useBreakpoints` and scss utilities (`atMediaUp`) are mobile-first, meaning `min-width` based. We write CSS for the smallest possible viewport first, then use media queries to target specific larger sizes.

This PR updates `Table` and `useBreakpoints` to reflect this, where `"none"` is a possible media match (expected for screens smaller than `s`).

**tl;dr** This PR allows `Table` to target viewports to the left of `s` in the illustrations below

Screenshot 2025-07-14 at 1 57 32 PM
Screenshot 2025-07-14 at 1 57 37 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,7 +85,7 @@
         "react-dom": "^18.3.1",
         "sass": "^1.45.0",
         "sass-loader": "^12.1.0",
-        "semantic-release": "^19.0.0",
+        "semantic-release": "^24.2.6",
         "storybook": "^8.6.0-alpha.0",
         "style-dictionary": "^3.9.2",
         "style-loader": "^0.23.1",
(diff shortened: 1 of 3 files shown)


src/hooks/useBreakpoints/index.jsx
```diff
@@ -17,13 +17,22 @@ const useBreakpoints = (
...
   const xl = useMediaQuery(queries.xl);
-  return { s, m, l, xl };
+
+  const largestSatisfiedBreakpoint =
+    [
+      { name: "xl", satisfied: xl },
+      { name: "l", satisfied: l },
+      { name: "m", satisfied: m },
+      { name: "s", satisfied: s },
+    ].find((bp) => bp.satisfied)?.name ?? "none";
```
akdetrick, line 33: #### Why do we need this?

Until now, the hook only returned booleans for each breakpoint. There was no way to get any idea of the "current" breakpoint. Because we use `min-width`/mobile-first breakpoints in NDS, `m` is still true when `l` is.

Our beta `Table` component needs this to match incoming props to breakpoint names.

src/Table/util/breakpoint.ts
```diff
@@ -1,42 +1,23 @@
-import type { ColMinBreakpoint } from "..";
-
-const breakpointOrder = ["s", "m", "l"] as const;
-type OrderedBreakpoint = (typeof breakpointOrder)[number];
+import type { ColMinBreakpoint, ViewportBreakpoint } from "..";
```
akdetrick, line 1: This change is just style cleanup and adding new test cases

src/Table/util/breakpoint.ts
```diff
@@ -1,42 +1,23 @@
...
-  }
-  return currentBreakpoint;
-};
+export const isBreakpointSatisfied = (
+  minRequired: ColMinBreakpoint,
+  current: ViewportBreakpoint,
+): boolean =>
+  minRequired === "*"
+    ? true
+    : BREAKPOINT_ORDER[current] >= BREAKPOINT_ORDER[minRequired];
```
akdetrick, line 23: In the course of making other changes, I kept breaking the function of `*` in the `colVisibility` prop.
I added an additional test case in this branch and discovered we missed some behavior in this implementation.

josh-narmi (approved):

PR #1682, opened 2025-06-17
feat(MenuButton): support footer item in MenuButton

elliot-wilson:
Designs for the AI prompt call for a menu button with a footer:

image

This PR is one attempt at making this possible. It's probably not the _right_ way, so I'm open very open to feedback/suggestions. My goal was to show the desired look/behavior.

Technical details:
- Moves our handling of React-Aria `MenuItem` into its own file to make it reusable for both the main content and the footer.
- Adds logic to render a `footerItem` using `MenuItem` with small style adjustment (e.g. top border)
- Adds story

https://github.com/user-attachments/assets/b6bfb895-5c51-4cb0-a883-cbc5b8efa78e

### DIFF
--- a/src/MenuButton/AriaMenuItem.js
+++ b/src/MenuButton/AriaMenuItem.js
@@ -1,0 +1,78 @@
+import React from "react";
+import cc from "classcat";
+import PropTypes from "prop-types";
+import { MenuItem as AriaMenuItem } from "react-aria-components";
+import iconSelection from "src/icons/selection.json";
+import Row from "../Row";
+
+export const VALID_ICON_NAMES = iconSelection.icons.map(
+  (icon) => icon.properties.name,
+);
+
+const MenuItem = ({
+  id,
+  label,
+  startIcon,
+  endIcon,
+  roundedTop,
+  roundedBottom,
+  className,
+}) => (
+  <AriaMenuItem
+    key={id}
+    id={id}
+    value={id}
+    startIcon={startIcon}
+    endIcon={endIcon}
+    /**
+     * react-aria provides a className interface similar
+     * to render props
+     */
+    className={({ isSelected, isFocused, isDisabled }) =>
+      cc([
+        "nds-menubutton-item",
+        "padding--x--s padding--y--xs",
+        {
+          "nds-menubutton-item--highlighted": isSelected || isFocused,
+          "nds-menubutton-item--disabled": isDisabled,
+          "rounded--top": roundedTop,
+          "rounded--bottom": roundedBottom,
+          [className]: className,
+        },
+      ])
+    }
+  >
+    <Row gapSize="s">
+      {startIcon && (
+        <Row.Item shrink>
+          <span className={`narmi-icon-${startIcon}`} />
+        </Row.Item>
+      )}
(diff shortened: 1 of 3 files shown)


src/MenuButton/index.js
```diff
@@ -35,9 +31,11 @@ const MenuButton = ({
   alignment = "start",
   offset = 2,
   children,
+  footerItem,
 }) => {
   const [isOpen, setIsOpen] = useState(false);
   const menuItems = React.Children.toArray(children);
+  const allItems = footerItem ? menuItems.concat(footerItem) : menuItems;
```
akdetrick, line 38: For the subcomponent interface it would be more like...

```jsx
const allItems = React.Children.toArray(children);
const menuItems = allItems.filter(isMenuItem); // see implementation of filter in other downshift components
const footerItem = allItems.find(isFooterItem); // take the first MenuButton.Footer in children and use that
```

I don't feel strongly about this though. Using `footerItem` might be more consistent with the model we use in other components with a pinned footer, e.g. `Dialog`

src/MenuButton/AriaMenuItem.js
```diff
@@ -0,0 +1,78 @@
...
+import PropTypes from "prop-types";
+import { MenuItem as AriaMenuItem } from "react-aria-components";
+import iconSelection from "src/icons/selection.json";
+import Row from "../Row";
+
+export const VALID_ICON_NAMES = iconSelection.icons.map(
+  (icon) => icon.properties.name,
+);
+
+const MenuItem = ({
```
akdetrick, line 12: **optional suggestion**

Rename to `AriaMenuItem`, even though the default export makes it arbitrary.

akdetrick (commented):
Looks good! The footer item is managed as a selectable item within the downshift hook, so it's selectable by keyboard as expected.

Extracting `MenuItem` into `AriaMenuItem` is a good improvement in the main render function

## Questions
- What does this look like when you put enough items in MenuButton to cause the list area above the footer to overflow?
- I left a suggestion to implement as`MenuButton.Footer`, but I think that the `footerItem` prop has it's merits as an interface, too. I'll leave it up to you

src/MenuButton/AriaMenuItem.js
elliot-wilson: done!

elliot-wilson:
> * What does this look like when you put enough items in MenuButton to cause the list area above the footer to overflow?

Even before this change, it doesn't look like `MenuButton` supported overflow scroll.

To your point, though, this technically isn't "footer" (instead of having a fixed position while the rest of the items scroll), but instead it's just a guaranteed final `MenuButton.Item` with different styling.

image

I tried playing around with making this a proper footer, e.g. by using the same pattern from `AutoCompleteModal`:

https://github.com/narmi/design_system/blob/0cfb0db7b39dc9ffd64ebe5ef6e3251d45d40a6a/src/AutocompleteModal/AutoComplete.tsx#L120-L124

But because we using slightly different patterns in each (e.g. ARIA MenuItem + downshift useLayer in MenuButton, our own item CSS + downshift useCombobox in AutoCompleteModal), it's not straightforward to me.

So I guess: Do you think that this is a blocker for this PR?

PR #1692, opened 2025-06-18
feat(ContentCard): add disabled prop

akdetrick:
Closes NDS-1595

Back in April, [I introduced the `error` prop to `ContentCard`](https://github.com/narmi/design_system/pull/1608) and **assumed** we wanted to disable a `button` or `toggle` kind card when it's in an error state.

The assumption was incorrect. This PR:

- Adds a new, separate control to disable a ContentCard
- Removes cursor rule and disabling from the error state
- Disables the hover/press styles when a card is in error state

### DIFF
--- a/src/ContentCard/index.tsx
+++ b/src/ContentCard/index.tsx
@@ -49,8 +49,13 @@ interface ContentCardProps {
   testId?: string;
   /**
    * Error state for `toggle` and `button` variants
+   * By default, this triggers `disabled` to be true - you may override the disabled property to `false`.
    */
   error?: string;
+  /**
+   * Disables the card, preventing any click events.
+   */
+  disabled?: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/ContentCard/index.tsx
```diff
@@ -49,8 +49,13 @@ interface ContentCardProps {
   testId?: string;
   /**
    * Error state for `toggle` and `button` variants
+   * By default, this triggers `disabled` to be true - you may override the disabled property to `false`.
```
akdetrick, line 52:
```suggestion.
```

akdetrick, line 52:
```suggestion
```

src/ContentCard/index.tsx
```diff
@@ -51,6 +51,10 @@ interface ContentCardProps {
    * Error state for `toggle` and `button` variants
    */
   error?: string;
+  /**
+   * Disables the card, preventing any click events.
+   */
+  disabled?: boolean;
```
akdetrick, line 57: It's notable that design does have a concept of "disabled" for a ContentCard, but it's accomplished in code by composing ContentCard with a [DisabledShim](https://narmi.github.io/design_system/?path=/docs/components-disabledshim--docs)

PR #1698, opened 2025-06-23
Combobox: fix `onSelectedItem` handler for user-cleared inputs

akdetrick:
Closes NDS-1628

Updates `onSelectedItem` downshift event handler to call the `props.onChange` function of Combobox with an empty string (instead of erroring) when a user clears the value of the input.


xiexiewenting (approved):
thank you!!

https://github.com/user-attachments/assets/268ff6f4-fe20-4067-857b-5003cb8ef674

PR #1771, opened 2025-07-17
feat(AutocompleteModal): allow use of searchValue prop for filtering items

elliot-wilson:
I noticed that the `AutocompleteModalItem` component accepts a `searchValue` prop, but it wasn't being used anywhere:

https://github.com/narmi/design_system/blob/d2ae373c5ea59ff39507d5070d0824671201477f/src/AutocompleteModal/Item.tsx#L17-L22

I assume this was just an oversight?

In any case, I want to be able to distinguish between the value passed into the `AutocompleteModal`'s `onChange` handler and the typeahead value used for searching. The classic scenario for this is searching for users by display name but using the selected user's UUID to make some API call, which is what I depict in the story I added.

As a drive-by: two typo fixes.

### DIFF
--- a/src/AutocompleteModal/AutoComplete.tsx
+++ b/src/AutocompleteModal/AutoComplete.tsx
@@ -17,11 +17,12 @@ export interface AutoCompleteProps {
   onChange?: (item: AutoCompleteItem) => void;
   /** Input change event callback */
   onInputChange?: (e: string) => void;
-  /** Optional pinned footer conent; use for action buttons */
+  /** Optional pinned footer content; use for action buttons */
   footerContent?: React.ReactNode;
 }
 
-export const itemToString = (item: AutoCompleteItem) => item.props.value || "";
+export const itemToString = (item: AutoCompleteItem) =>
+  item.props.searchValue || item.props.value || "";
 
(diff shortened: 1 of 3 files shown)


src/AutocompleteModal/AutoComplete.tsx
```diff
@@ -17,11 +17,12 @@ export interface AutoCompleteProps {
...
   /** Input change event callback */
   onInputChange?: (e: string) => void;
-  /** Optional pinned footer conent; use for action buttons */
+  /** Optional pinned footer content; use for action buttons */
   footerContent?: React.ReactNode;
 }
 
-export const itemToString = (item: AutoCompleteItem) => item.props.value || "";
+export const itemToString = (item: AutoCompleteItem) =>
+  item.props.searchValue || item.props.value || "";
```
elliot-wilson, line 25: This is the same pattern we use in a bunch of other places, e.g.

https://github.com/narmi/design_system/blob/d2ae373c5ea59ff39507d5070d0824671201477f/src/Select/index.js#L109-L112

https://github.com/narmi/design_system/blob/d2ae373c5ea59ff39507d5070d0824671201477f/src/MultiSelect/index.js#L20-L21

etc

PR #1809, opened 2025-08-04
fix(DropdownTrigger): prevent width overflow by allowing value to wrap

akdetrick:
Closes NDS-1734

The original bug report noted that when `AccountSelector` in banking is used in an NDS dialog, it pushes the width of the container past the edge of the viewport on mobile sizes. The `AccountSelector` implements `Select`, which uses `DropdownTrigger` to mimic an input.

I was able to reproduce the bug in isolation in NDS. The fix is to remove text overflow rules that weren't actually working and allow the value of a dropdown trigger to wrap if necessary.

### Before
Screenshot 2025-08-04 at 5 38 34 PM

### After
(red outline for debugging)
Screenshot 2025-08-04 at 5 40 38 PM

### DIFF
--- a/src/DropdownTrigger/index.scss
+++ b/src/DropdownTrigger/index.scss
@@ -1,6 +1,7 @@
 .nds-dropdownTrigger {
   position: relative;
   min-height: rem(48px);
+  max-width: 100%;
   display: flex;
   align-items: center;
 }
@@ -45,9 +46,6 @@
 .nds-dropdownTrigger-value {
   color: var(--font-color-primary);
   max-width: 100%;
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
   padding-right: rem(24px);
 }
 


src/DropdownTrigger/index.scss
```diff
@@ -1,6 +1,7 @@
 .nds-dropdownTrigger {
   position: relative;
   min-height: rem(48px);
+  max-width: 100%;
```
akdetrick, line 4: just to be safe

src/DropdownTrigger/index.scss
```diff
@@ -45,9 +46,6 @@
 .nds-dropdownTrigger-value {
   color: var(--font-color-primary);
   max-width: 100%;
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
```
akdetrick, line 50: 1. This wasn't actually working as expected
2. It's best not to truncate the value we render when an item is selected, e.g. two accounts with the same start words in their name

XihaoLuo (approved):
Makes sense.

PR #1816, opened 2025-08-06
fix(CollapsibleCard): prevent collapse on content clicks

akdetrick:
Closes NDS-1754

Ensures click handlers for "header trigger mode" (when the `trigger` prop is `"header"`) are only attached to the header instead of the entire card.

Prevents content cards closing unexpectedly on content clicks.


wrbrand (approved):
Does what it says on the tin.

PR #1822, opened 2025-08-06
feat(TableInput): add TableInput component for editable cells

akdetrick:
Closes NDS-1755

Add `TableInput` component for use in `Table kind="editable"`.

See storybook preview on this PR for more details.

Screenshot 2025-08-06 at 5 52 32 PM


src/Table/index.scss
```diff
@@ -23,10 +23,13 @@
   grid-template-columns: subgrid;
   grid-column: 1 / -1; // participate in all parent columns
   align-items: center;
+  padding: 0 rem(8px);
```
akdetrick, line 26: the left and right side of rows needed a bit more space according to figma

src/TableInput/index.stories.tsx
```diff
@@ -0,0 +1,107 @@
+import React, { useState } from "react";
+import TableInput from "./";
+import Table from "../Table";
+// Import the TableInputProps type for type safety
```
akdetrick, line 4:
```suggestion
```

src/TableInput/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  isDisabled = false,
+  placeholder,
+}: TableInputProps) => {
+  return (
+    <div className="nds-tableField-input">
+      <input
+        type="text"
+        aria-label={label}
+        value={value}
+        onChange={(e) => onChange(e.target.value)}
```
akdetrick, line 33: **Change needed**

We need this to be as composable as `TextInput`, so we should not put this facade around the native html dom event `e` and instead type this for a native event

blooies (approved):
played around with it on storybook, looks good!

PR #1832, opened 2025-08-13
feat(TableAutocomplete): add autocomplete input for tables

akdetrick:
closes NDS-157

Adds `TableAutocomplete` input.

Composes `TableInput` with a dropdown menu using `useCombobox` downshift hook for state and event management with `react-laag` (`useLayer`) for positioning of the dropdown.

#### Listening for selected items (uncontrolled)
We're typically only interested in the `value` once an item is actually selected via the `onChange` event:

`<TableAutocomplete onChange={/* handle value of selected item */} .../>`

Here we don't need to worry about what's going on in the `input` internally, only the item selection.

#### Setting a default (controlled)

Pass a string matching the `value` of one of the autocomplete items to `inputValue`. When passing an `inputValue`, the component becomes fully controlled and 

`const [val, setVal] = useState("someItemValue")`
`<TableAutocomplete inputValue={val} onInputChange={({ target}) => setVal(target.value) ... />`

--------

Screenshot 2025-08-13 at 6 07 03 PM
Screenshot 2025-08-13 at 6 07 09 PM
Screenshot 2025-08-13 at 6 07 14 PM
Screenshot 2025-08-13 at 6 07 31 PM

### DIFF
--- a/src/TableAutocomplete/Item.tsx
+++ b/src/TableAutocomplete/Item.tsx
@@ -1,0 +1,21 @@
+import React from "react";
+
+export interface TableAutocompleteItemProps {
+  /**
+   * String representation of the option.
+   */
+  value: string;
+  /**
+   * Custom typeahead string. By default, typeahead uses `value`.
+   */
+  searchValue?: string;
+  children?: React.ReactNode;
+}
+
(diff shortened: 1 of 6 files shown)


blooies (approved):
looking good!!

PR #1843, opened 2025-08-20
fix(Button): prevent small buttons from scaling on mobile

akdetrick:
Closes NDS-1746

Allow only the `m` button size to scale its height on mobile viewports. 

### Before
Screenshot 2025-08-20 at 4 21 42 PM

### After
Screenshot 2025-08-20 at 4 22 25 PM

### DIFF
--- a/src/Button/sizes.scss
+++ b/src/Button/sizes.scss
@@ -29,4 +29,8 @@
     margin: rem(8px) rem(32px);
     max-width: rem(240px);
   }
+  // Medium is our default size. As such, we want it to scale up on smaller viewports.
+  @media (max-width: #{map.get($breakpoints, "s")}) {
+    min-height: rem(48px) !important;
+  }
 }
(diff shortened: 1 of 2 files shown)


src/Button/sizes.scss
```diff
@@ -29,4 +29,8 @@
     margin: rem(8px) rem(32px);
     max-width: rem(240px);
   }
+  // Medium is our default size. As such, we want it to scale up on smaller viewports.
+  @media (max-width: #{map.get($breakpoints, "s")}) {
+    min-height: rem(48px) !important;
+  }
```
akdetrick, line 35: This rule now belongs to `%btn-size-m`, and that size only.

XihaoLuo (approved):
Makes sense.

src/Button/sizes.scss
XihaoLuo: Seeing https://linear.app/narmi/issue/NDS-1746/should-the-nds-button-height-vary-depending-on-size-prop-in-mobile#comment-34343931 

Don't we want to leave existing behavior for size `l` as well?

PR #1844, opened 2025-08-21
feat(button): add isSpread prop

agreatscott:
There's a design need to have a button with an end icon, where the two are split across the length of the button (think space-between). This PR attempts to implement that in the current Button component. 

I don't love the implementation, and it's a bit messy, but it does the trick. Open to thoughts/comments. 

Desired from Figma:
image

Actual storybook with this change
image

### DIFF
--- a/.github/workflows/chromatic-main.yaml
+++ b/.github/workflows/chromatic-main.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout main branch
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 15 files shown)


akdetrick:
Biggest concerns:
- `startContent`, `startIcon`, `endContent`, `endIcon` may need different flex rules
- `SplitButton` will also take this prop. What does a spread splitButton look like? We probably want to pack all button internal items and make the secondary button the shrink member of the flex

src/Button/index.scss
```diff
@@ -97,3 +97,13 @@
...
 }
+
+.nds-button--spread {
+  justify-content: stretch !important;
+
+  &.nds-button--xs .nds-button-content,
+  &.nds-button--s .nds-button-content,
+  &.nds-button--m .nds-button-content {
+    width: 100% !important;
+  }
```
akdetrick, line 108: I think what we want to happen with `isSpread` is for the root element (the button or anchor) takes up 100% of parent width
```suggestion
    display: flex; /* instead of inline flex , makes it behave like display: block */
  }
```

PR #1847, opened 2025-08-25
feat(Pagination): use sliding window behavior

akdetrick:
Closes NDS-1768

Refactor core pagination logic into a react hook; uses a more standard UX behavior matching this example: <https://atlassian.design/components/pagination/examples>

- Remove `_getAttributes` helper function
- Add `usePagination` to calculate attributes needed for rendering
- Add tests for `usePagination`

### Why not use an existing hook?
There aren't many libraries that expose pagination logic in a convenient hook, and most of them manage the `onPageChange` responsibility. It was frankly easier to roll my own using material and atlassian as implementation references.

### Anatomy of `Pagination`
These are the parts that usePagination is managing...

**Red:** the "sliding window" aka `visiblePages`
**Blue:** the ellipsis that appear when the first or last page is outside the visible window respectively
**Lime:** first and last pages
**Purple:** next/prev buttons
Screenshot 2025-08-25 at 6 56 39 PM

### Preview build (updated 9/3)
[View Storybook preview](https://60620d422ffdf100216415b2-vbhgdbdyih.chromatic.com/)

### DIFF
--- a/src/Pagination/index.js
+++ b/src/Pagination/index.js
@@ -1,7 +1,6 @@
 import React, { useState } from "react";
 import PropTypes from "prop-types";
 import cc from "classcat";
-import Row from "../Row";
 import { usePagination } from "./usePagination";
 
 const noop = () => {};
@@ -70,8 +69,8 @@ const Pagination = ({
   const pagination = (
     <div className="nds-typography nds-pagination" data-testid={testId}>
       <nav aria-label="pagination">
-        <Row gapSize="xxs" alignItems="center" as="ul">
-          <Row.Item as="li" shrink>
+        <ul>
+          <li>
             <span
               role="button"
               tabIndex={0}
@@ -93,10 +92,10 @@ const Pagination = ({
             >
               <i role="img" className="narmi-icon-chevron-left fontSize--l"></i>
             </span>
-          </Row.Item>
+          </li>
 
           {showFirstPage && (
-            <Row.Item as="li" shrink>
+            <li>
               <span
                 role="button"
                 tabIndex={0}
@@ -112,16 +111,16 @@ const Pagination = ({
               >
                 1
               </span>
-            </Row.Item>
+            </li>
           )}
           {showFirstPage && (
-            <Row.Item as="li" shrink>
+            <li>
               <div className="nds-pagination-ellipsis">&hellip;</div>
-            </Row.Item>
+            </li>
           )}
 
           {visiblePages.map((page, i) => (
-            <Row.Item as="li" key={page} shrink>
+            <li key={`page-${page}`}>
               <span
                 role="button"
                 tabIndex={0}
@@ -143,16 +142,16 @@ const Pagination = ({
               >
                 {page.toString()}
               </span>
-            </Row.Item>
+            </li>
           ))}
 
           {showLastPage && (
-            <Row.Item as="li" shrink>
+            <li>
               <div className="nds-pagination-ellipsis">&hellip;</div>
-            </Row.Item>
+            </li>
           )}
           {showLastPage && (
-            <Row.Item as="li" shrink>
+            <li>
               <span
                 role="button"
                 tabIndex={0}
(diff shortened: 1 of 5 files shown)


src/Pagination/usePagination.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    it("shows all pages when they fit in window", () => {
+      const { result } = renderHook(() => usePagination({ totalPages: 5 }));
+      expect(result.current.visiblePages).toEqual([1, 2, 3, 4, 5]);
+    });
+
+    it("centers selected page when possible", () => {
+      const { result } = renderHook(() =>
+        usePagination({ totalPages: 10, selectedPageNumber: 6 }),
+      );
+      expect(result.current.visiblePages).toEqual([5, 6, 7]);
```
akdetrick, line 15: The default `windowSize` is `5`, but we adjust it according to how many ellipsis are showing. With 6 selected out of 10 pages, there will be ellipsis on both sides, resulting in 3 visible pages (5 - 2 ellipsis) with the `6` centered.

src/Pagination/index.js
```diff
@@ -1,61 +1,10 @@
...
- */
-export const _getAttributes = (
-  totalPages,
-  selectedPage,
-  segmentSize = MAX_VISIBLE_PAGES
-) => {
-  // create a list of page segments
-  // each segment is an array of page numbers.
-  const pageSegments = [...new Array(totalPages)]
-    .map((p, i) => i + 1) // [1,2,3,4,5,6,7, ...]
```
akdetrick, line 25: You'll notice this trick being used in my `range` implementation in the usePagination hook.
It's wild that JS still doesn't have native range

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file. See
 [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 4.38.0 (2025-08-21)
```
akdetrick, line 6: ? - need to check git history on this one

akdetrick:
In draft to address a bad rebase.

src/Pagination/index.scss
```diff
@@ -1,18 +1,30 @@
...
+    display: grid;
+    grid-auto-flow: column;
+    justify-items: center;
+    align-items: center;
+    gap: var(--space-xxs);
+
+    // Every column should be the same width as the widest item to prevent
+    // UX shifting when paging between single, double, tripple, (or more)
+    grid-auto-columns: 1fr;
+    width: max-content;
```
akdetrick, line 26: For every column created for an item, make it `1fr` of total columns, which divides the available space equally. Each column is set to max-width, so the widest item will participate in the meaning of `1fr`. This is a trait unique to CSS grid

src/Pagination/usePagination.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (start < 1) {
+    start = 1;
+    end = effectiveSize;
+  } else if (end > totalPages) {
+    end = totalPages;
+    start = totalPages - effectiveSize + 1;
+  }
+
+  return range(start, end);
+};
```
josh-narmi, line 70: Nit: the computation here isn't crazy, but since the function is pure, wrapping the whole thing in a `useMemo()` would improve performance.

src/Pagination/usePagination.ts
```diff
@@ -0,0 +1,102 @@
...
+  return {
+    visiblePages,
+    selectedPage: normalizedSelectedPage,
+    selectedIndex: visiblePages.indexOf(normalizedSelectedPage),
+    showFirstPage: visiblePages[0] > 1,
+    showLastPage: visiblePages[visiblePages.length - 1] < normalizedTotalPages,
+    showPrev: normalizedSelectedPage > 1,
+    showNext: normalizedSelectedPage < normalizedTotalPages,
+  };
+};
```
josh-narmi, line 102: Nit: Same comment about memoizing

josh-narmi (approved):
This is so cool! Left a nit comment about memoization

PR #1853, opened 2025-09-03
feat(Dialog): adapt to all viewport orientations

akdetrick:
Closes NDS-1807

Fixes Dialog showing as the desktop style modal when mobile devices are in landscape orientation.

### Changes
- Add optional `includeHeight` argument to NDS's internal `atMediaUp` mixin
- Update mixin calls in Dialog scss
- Use `clamp` on the top padding we use to space Dialog from the top of the viewport. I've never liked how we push it to the bottom on shorter viewports, so this ensures we place it closer to the center of the screen with a min gutter of default space and a max of the original `120px`.

### Why the bug happened
Media queries generated from the `atMediaUp(<size>)` mixin are mobile-first, meaning based on min values. We currently only use `min-width`, which works well for most things.

As for the Dialog, we intentionally take over the full viewport on smaller devices, but when you turn a device sideways, it might be wider than the width our `s` breakpoint queries.

### The fix
Pass a special arg to `atMediaUp` calls in Dialog scss, so we can verify both the height and width when switching from a full viewport takeover to a modal centered dialog at larger sizes.

### Testing / Verification
Open up your responsive view in your web inspector and try out the "Scrolling content" Dialog story. 
Use the "view in a new tab" feature to make the story full page (top right of storybook):
Screenshot 2025-09-03 at 5 19 19 PM

You'll see that regardless of the size or aspect ratio, the takeover display of the modal is triggered whenever there isn't adequate space for the centered modal dialog:

https://github.com/user-attachments/assets/39a26285-27c5-4345-9513-9c9a85a8ac15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,16 @@
 All notable changes to this project will be documented in this file. See
 [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## <small>4.39.1 (2025-09-04)</small>
+
+* fix(MenuButton): limit height of popover to 60vh ([ff0dfaef4bd294c2c458984c63113d172894ead2](https://github.com/narmi/design_system/commit/ff0dfaef4bd294c2c458984c63113d172894ead2))
+
+## 4.39.0 (2025-09-03)
+
+* feat(Pagination): use sliding window behavior ([48c4f162239530074921f41c80b4ed41f283ce60](https://github.com/narmi/design_system/commit/48c4f162239530074921f41c80b4ed41f283ce60))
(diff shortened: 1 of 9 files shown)


src/Dialog/index.scss
```diff
@@ -18,10 +18,10 @@
   z-index: 3;
 
   // create spaces around viewport edges for tablet or larger
-  @include atMediaUp("s") {
+  @include atMediaUp("s", $includeHeight: true) {
     align-items: flex-start;
     padding: var(--space-default);
-    padding-top: 120px;
+    padding-top: clamp(var(--space-default), 10vh, 120px);
```
akdetrick, line 24: `clamp(minValue, preferredValue, maxValue)`

PR #1864, opened 2025-09-16
feat(hooks): add momentum scroll helpers

akdetrick:
Closes NDS-1798

An FI recently observed a bug on iOS Safari, where scrolling while a dropdown is open will cause the dropdown to float away from the trigger and/or be positioned incorrectly.

This was observed in CAO/BAO _before_ we replaced the Combobox usage with Select, which made matters worse because iOS Safari will auto scroll to a focused input, which was messing up the positioning by the scroll difference.

### The fix
By mounting the popovers in Combobox and Select to `body` with `fixed` positioning, the dropdown is allowed to scroll with the document without trying to reposition itself.

- add hook to detect when on a touch device
- update the `useLayer` hook options for positioning dropdowns in `Select` and `Combobx` ONLY when on a touch device

## Example
Note the auto-scrolling that happens on input focus of the Combobox here. Using fixed position helps us avoid Safari doing scroll magic after we've already calculated the correct dropdown position. The dropdown may now scroll with body:
Sep-16-2025 14-35-10


akdetrick:
Not doing it this way. Something is off when either Select or Combobox is in a Dialog in this branch.

Closing in favor of trying a native CSS solution with a react-laag JS fallback

PR #1866, opened 2025-09-21
feat(Select,Combobox): use progressive dropdown positioning

akdetrick:
Closes NDS-1798

- Add new hooks for progressively enhanced dropdown layout helpers
- Update `Combobox` and `Select` to use our new helpers

This PR primarily fixes an iOS Safari issue where Combobox input focus causes an auto-scroll, making the original positioning engine lose its position. This PR introduces a fallback for Safari 26 and below that uses the [Visual Viewport API](https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API) to set the dropdown via fixed positioning.

For more modern browsers, we feature detect for [CSS Anchor Positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning) and apply the CSS-only alternative when available. 

I would guess that about 60% of our users have a browser that supports anchor positioning. Whenever someone upgrades to a newer browser, they'll get it for free.

### What is Anchor positioning?
You can declare an element as a positioning anchor by applying the `anchor-name` property, which takes a custom property value like `--my-anchor`. That custom property may be referenced by other elements to place them in anchor position against `--my-anchor`. 

**Example:** https://codepen.io/akdetrick/pen/ByjBNwN

### What is the Visual Viewport API
Instead of `window.<dimension>`, I use`visualViewport.<dimension>`. That's basically it. The visual viewport represents only the part of the document that is visible within the viewport, as opposed to the whole document. It was specifically invented to solve Apple's refusal to respect web standards with their auto-scroll on focus behavior.

### With anchor positioning
In most modern browsers, the CSS-only solution is applied:
with_anchor_pos

### With fallback positioning
The polyfill hook detects if anchor positioning is not supported, in which case, it uses some JS to determine what the `top` and `left` values for a fixed position dropdown should be. This is where we use the visual viewport to work around Safari.
Safari 26 and up support anchor positioning.

The flip isn't very smooth via browserstack, but is acceptable on a real device compared to the original behavior.
safari_scroll
safari_flilp


src/Select/index.test.js
```diff
@@ -99,11 +99,20 @@ describe("Select", () => {
...
         <Select.Item value="savings"></Select.Item>
-      </Select>
+      </Select>,
     );
 
     expect(screen.getByText("Account Type")).toBeInTheDocument();
-    expect(screen.getByRole("listbox")).toBeEmptyDOMElement();
+
+    // The listbox is always in the DOM but hidden when closed.
+    // We can't use getByRole when display: none because it's removed from the accessibility tree
```
akdetrick, line 108:
```suggestion
```

src/Combobox/index.test.js
```diff
@@ -110,7 +110,12 @@ describe("Combobox", () => {
   it("renders as expected with basic props", () => {
     render(<Combobox label={LABEL}>{STATE_ITEMS}</Combobox>);
     expect(screen.getByText(LABEL)).toBeInTheDocument();
-    expect(screen.getByRole("listbox")).toBeEmptyDOMElement();
+
+    // The listbox is always in the DOM but hidden when closed.
+    // We can't use getByRole when display: none because it's removed from the accessibility tree
```
akdetrick, line 115:
```suggestion
```

src/hooks/useDropdownLayer/useAnchorPolyfill.ts
```diff
@@ -0,0 +1,115 @@
...
+  isOpen: boolean;
+}
+
+/**
+ * Polyfill for CSS anchor positioning dropdown behavior.
+ * If anchor positioning is supported, no element observers will be attached.
+ *
+ * This polyfill uses fixed positioning with JavaScript flip detection to position
+ * dropdowns relative to their anchor elements. It handles intelligent positioning
+ * logic (showing above vs below based on available space).
```
akdetrick, line 21:
```suggestion
```

src/Combobox/index.scss
```diff
@@ -1,42 +1,10 @@
-.nds-combobox {
-  position: relative;
-}
-
```
akdetrick, line 4: This was irrelevant in the first place - the dropdown is portalled outside of the nds-combobox element

src/Combobox/index.scss
```diff
@@ -1,42 +1,10 @@
...
-  &--error {
-    border: 1px solid var(--color-errorDark);
-  }
-
-  &--bottom {
-    border-top: 1px solid var(--border-color-default);
-    border-top-right-radius: 0;
-    border-top-left-radius: 0;
-  }
+  @extend %nds-dropdown-layer;
```
akdetrick, line 6: **Style changes:**

Shared styles between `Select` and `Combobox` have been extracted into the `%nds-dropdown-layer` sass placeholder.

State and directional variants like `--top` or `--active` are no longer necessary.

src/base-styles/scss-utils.scss
```diff
@@ -71,3 +71,25 @@ $breakpoints: (
...
+* Dropdown Placeholder
+* Apply to element that holds the dropdown `listbox`/`ul`
+*/
+%nds-dropdown-layer {
+  max-height: 50vh;
+  overflow-y: auto;
+  border: 1px solid var(--theme-primary);
+  border-radius: var(--border-radius-default);
+  margin-top: var(--space-xxs) !important;
+  margin-bottom: var(--space-xxs) !important;
```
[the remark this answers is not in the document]
akdetrick: **Required**

Fix this. We should only manage the gap in one place.

josh-narmi (approved):
Awesome stuff!

PR #1876, opened 2025-09-30
fix(Table.Row): Add border radius to `only-of-type` rows

josh-narmi:
- [NDS-1839: Add radius to corners of table row](https://linear.app/narmi/issue/NDS-1839/add-radius-to-corners-of-table-row)

https://github.com/user-attachments/assets/95ab1608-2808-4c7e-bd96-4aa59e67906e

### DIFF
--- a/src/Table/index.scss
+++ b/src/Table/index.scss
@@ -43,6 +43,9 @@
   &:last-of-type {
     border-radius: 0 0 var(--border-radius-default) var(--border-radius-default);
   }
+  &:only-of-type {
+    border-radius: var(--border-radius-default);
+  }
 
   // Clickable rows
   &--interactive:hover,


src/Table/index.scss
```diff
@@ -43,6 +43,9 @@
   &:last-of-type {
     border-radius: 0 0 var(--border-radius-default) var(--border-radius-default);
   }
+  &:only-of-type {
+    border-radius: var(--border-radius-default);
+  }
```
josh-narmi, line 48: With a single row, both `first-of-type` and `last-of-type` selectors were being triggered sequentially, leaving the row with only a bottom radius. The `only-of-type` selector resolves that issue.

PR #2060, opened 2026-04-22
fix(Alert): use alert icon for error kind instead of rotating

elliot-wilson:
Hilariously, we used to just vertically rotate the `info` icon for error Alerts instead of using the `alert-circle` icon. (It truly blows my mind that the `!` is just an upside-down `i`.) The problem was that if you supplied a custom icon, we'd rotate that, too:

https://github.com/user-attachments/assets/eaa0efcc-e45f-4658-b520-04449879f1e1

The fix is to remove the rotation and just use the correct icon.

https://github.com/user-attachments/assets/790fb6b4-9c7c-496e-891a-7aa6785d1432

### DIFF
--- a/src/Alert/index.scss
+++ b/src/Alert/index.scss
@@ -9,12 +9,6 @@
 .nds-alert--error {
   background-color: var(--color-errorLight);
 
-  // Rotating the info icon until we have a "!" icon
-  .nds-alert-icon {
-    display: inline-block;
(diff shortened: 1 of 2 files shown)


akdetrick (approved):

akdetrick:
I'll update visual regression baselines and document the process better while I'm at it

PR #2063, opened 2026-04-23
fix(Button): use readable color for disabled plain button

akdetrick:
Closes NDS-2573

Fixes plain button so the disabled state uses secondary text instead of `hint`.


akdetrick:
Author-merging for a one line change.

PR #2088, opened 2026-05-21
UI refresh/patch 1

akdetrick:
Fixes NDS-2708
Fixes NDS-2704
Fixes NDS-2701

Will be backported to v6.11.1 as `6.11.1-patch.1`.

- Rounds our `DisabledShim` to prevent noticeable overlap over rounded corners
- Fix timers for `Tooltip`
- Prevent anchor positioned containers from inheriting breaking properties until we introduce cascade layers everywhere

### DIFF
--- a/src/DisabledShim/index.scss
+++ b/src/DisabledShim/index.scss
@@ -8,10 +8,8 @@
   user-select: none;
   pointer-events: none;
   position: absolute;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
+  inset: 0;
   background-color: var(--bgColor-scrimLight);
+  border-radius: var(--border-radius-default);
   z-index: 2;
 }
--- a/src/Tooltip/index.stories.js
+++ b/src/Tooltip/index.stories.js
@@ -168,6 +168,58 @@ ArrowTracking.parameters = {
   },
 };
 
+export const MultipleTooltips = () => (
+  <div
+    style={{
+      height: "400px",
+      display: "flex",
+      flexDirection: "column",
+      justifyContent: "center",
+      alignItems: "center",
+      gap: "var(--space-xl)",
+    }}
+  >
+    <div style={{ display: "flex", gap: "var(--space-l)" }}>
+      <Tooltip text="Tooltip on top" side="top">
+        <Button>Top</Button>
+      </Tooltip>
+      <Tooltip text="Tooltip on the right" side="right">
+        <Button>Right</Button>
+      </Tooltip>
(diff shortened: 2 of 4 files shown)


src/base-styles/scss-utils.scss
```diff
@@ -82,6 +82,21 @@ $breakpoints: (
...
+  // <https://linear.app/narmi/issue/NDS-2718/breaking-require-consumers-to-use-a-cascade-layer>
+  display: block;
+  font-size: var(--font-size-default);
+  color: var(--font-color-default);
+  text-align: start;
+  text-transform: none;
+  letter-spacing: normal;
+  line-height: var(--line-height-default);
+  padding: 0;
+
```
akdetrick, line 99: Unfortunately cascade layers "demote" rules in cascade resolution compared to unlayered styles.

There is a long-term fix where we surround each app's CSS with `@layer app-name {...}` so we can then define NDS as its own cascade layer.

src/DisabledShim/index.scss
```diff
@@ -8,10 +8,8 @@
   user-select: none;
   pointer-events: none;
   position: absolute;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
+  inset: 0;
   background-color: var(--bgColor-scrimLight);
+  border-radius: var(--border-radius-default);
```
akdetrick, line 13: The overlap was less noticeable will the old default radius. 

`inset` is just shorthand for the directional properties

src/Tooltip/index.stories.js
```diff
@@ -168,6 +168,58 @@ ArrowTracking.parameters = {
   },
 };
 
+export const MultipleTooltips = () => (
```
akdetrick, line 171: Test case to catch this in a future snapshot test with actions

src/Tooltip/index.tsx
```diff
@@ -40,18 +40,18 @@ const Tooltip = ({
     open: 500,
     close: 100,
   };
-  let activeTimer: ReturnType<typeof setTimeout>;
+  const activeTimer = useRef<ReturnType<typeof setTimeout>>();
```
akdetrick, line 43: A `ref` is stable between renders. This prevents one timer from cancelling another when more than one Tooltip is active.

src/Tooltip/index.stories.js
akdetrick: baselines updated, but we still need storybook action tests on these

josh-narmi (approved):

PR #2092, opened 2026-05-23
ci(release): isolate OIDC from dependency lifecycle scripts

chris-griffin:
## Summary

Hardens the release workflow by splitting it into two jobs so that OIDC credentials (`id-token: write`) are never available when dependency lifecycle scripts execute.

## Why

GitHub Actions OIDC tokens (`id-token: write`) were previously granted at the **workflow** level, meaning every step—including `npm ci` which runs arbitrary package lifecycle scripts—had access to mint OIDC tokens. A compromised or malicious dependency could abuse this to request tokens and publish artifacts or access cloud resources.

`id-token: write` can only be scoped at the **job** level (not step level), so the only way to prevent lifecycle scripts from accessing OIDC is to run them in a separate job that lacks the permission.

## Changes

- **`build` job** – runs `npm ci` (with lifecycle scripts) and `npm run build`, then uploads `dist/` as an artifact. This job has only `contents: read`—no OIDC.
- **`release` job** – depends on `build`, downloads the prebuilt `dist/` artifact, runs `npm ci --ignore-scripts` (so semantic-release and plugins are available but no lifecycle scripts execute), then runs `npx semantic-release`. This job has `id-token: write` for npm provenance.

## Provenance

`provenance: true` in `release.config.js` is preserved—the release job still has `id-token: write` so npm can generate OIDC-backed provenance attestations.

## Notes

- No changes to `release.config.js`, `package.json`, or `.npmrc`.
- All action references remain pinned to full commit SHAs.


akdetrick:
I'm not exactly sure what is keeping this PR in draft, but the changes make sense to me broadly.
Putting this on the UI Platform radar in an [upcoming sprint](https://linear.app/narmi/issue/NDS-2916/nds-pr-isolate-oidc)

akdetrick (approved):
Isolates permissions to only the step that needs write access

PR #2101, opened 2026-06-10
fix(Select,Combobox): force JS polyfill when in scroll containers

akdetrick:
Fixes NDS-2881

Safari made an implementation error with CSS anchor positioning, newly supported in os/iOS version 26.
This only happens when in scroll containers, so we unfortunately have to dumb this down a little until Safari addresses the issue.

- Adds another check to `isAnchorPositionSupported` to return `false` if the browser suffers from this specific anchor positioning bug.
- Allow components to opt-in. This is less of a problem for tooltips, which we can portal by default instead. The JS fallback for tooltips looks less than desirable.

Safari with the fix, using the JS polyfill:
Screenshot 2026-06-09 at 4 32 51 PM


josh-narmi (approved):

josh-narmi (approved):

src/hooks/useSupportsAnchorPositioning/index.ts
```diff
@@ -1,29 +1,77 @@
...
+/**
+ * Runtime check for Safari's specific scroll container bug.
+ *
+ * Creates 3 hidden elements (scroll container + anchor + positioned layer),
+ * measures the layer width, and cleans up. If the layer doesn't match the
+ * anchor's 100px width, the browser has the bug.
+ *
+ * Only called when cssChecksPass() is true (browser claims support).
+ */
+const detectScrollContainerBug = (): boolean => {
```
akdetrick, line 37: ### About the bug

I'm having a hard time finding any bugzilla issues that reference this exact use case. I'm considering filing a bug with a test case so we can track it and remove this later.

Specific to Safari in macOS/iOS >= 26, some anchor positioning CSS features fail when the anchor or the positioned element are in an `overflow: auto` container which creates a new scroll context.

#### Why are we checking width?

Safari technically does support `anchor-size(width)`. You can use it in most cases without any problem. In the case of the bug, I suspect Safari may lose its reference to the anchor by `anchor-name`.

This does not represent the entire bug as we observe it, but is a good "canary" that indicates we should fall back to the JS polyfill if the component has opted in.

PR #2111, opened 2026-06-17
fix(TextInput): only show char counter on focus or error

akdetrick:
Fixes NDS-2619

- reduce size of error text and char counter
- hide visibility of char counter unless input is focused or in error state


akdetrick:
Author-merging. This must be included in a patch version going to .5 Monday 6/22.

This fixes a bug introduced in .5 with UI refresh.

PR #2115, opened 2026-06-18
fix(Tooltip): portal tooltip always

akdetrick:
Closes NDS-2879

The modular dashboard has a few more adversarial styles where things are positioned and z-indexed for reasons not immediately clear to me. This causes `Tooltip` to appear under some other UI elements.

The best way to resolve this is to make sure the `Tooltip` is always portalled to `document.body` to resolve stacking context separately.

This option is stable; `TableSelect` has been setting this `useDropdownLayer` option for months now, as Tables/rows are often overflow hidden.


akdetrick:
Author merging, as this pattern is already used in MenuButton and TableSelect (see PR description)

PR #2138, opened 2026-07-06
chore: replace usehooks-ts with local hook implementations

chris-griffin:
## Summary
- Removes the `usehooks-ts` runtime dependency. Only two of its hooks were used across the codebase:
  - `useMediaQuery` — used by `useBreakpoints`
  - `useOnClickOutside` — used by `AnchoredDialog`
- Adds minimal local equivalents at `src/hooks/useMediaQuery` and `src/hooks/useOnClickOutside` (internal only — not added to the public export surface).

## Behavior parity
- `useMediaQuery` initializes with the current match state on first render and subscribes to `change` events, matching usehooks-ts defaults. SSR-safe (`false` when `window` is unavailable).
- `useOnClickOutside` listens for `mousedown` (the usehooks-ts default event) and keeps the latest handler in a ref so the listener isn't resubscribed every render.
- Also drops the `lodash.debounce` transitive from the runtime dependency tree.

## Testing
- `npx tsc --noEmit` — clean
- `npx vitest run` — 296 passed, 4 skipped

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,8 +119,7 @@
     "raf-schd": "^4.0.3",
     "react-aria-components": "^1.0.0",
     "react-focus-lock": "^2.9.4",
-    "react-transition-group": "^4.4.5",
-    "usehooks-ts": "^3.1.1"
+    "react-transition-group": "^4.4.5"
   },
   "overrides": {
     "tar": "7.5.10",
--- a/src/hooks/useMediaQuery/index.tsx
+++ b/src/hooks/useMediaQuery/index.tsx
@@ -1,0 +1,23 @@
+import { useState, useEffect } from "react";
+
+/**
+ * Tracks whether a CSS media query currently matches the viewport.
+ * SSR-safe: resolves to `false` when `window` is unavailable.
+ */
+const useMediaQuery = (query: string): boolean => {
+  const [matches, setMatches] = useState(
+    () => typeof window !== "undefined" && window.matchMedia(query).matches,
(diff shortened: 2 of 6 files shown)


src/hooks/useMediaQuery/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { useState, useEffect } from "react";
+
+/**
+ * Tracks whether a CSS media query currently matches the viewport.
+ * SSR-safe: resolves to `false` when `window` is unavailable.
+ */
+const useMediaQuery = (query: string): boolean => {
```
akdetrick, line 7: @copilot Do we need an observer to re-run the query if orientation or viewport changes?

package.json
```diff
@@ -119,8 +119,7 @@
     "raf-schd": "^4.0.3",
     "react-aria-components": "^1.0.0",
     "react-focus-lock": "^2.9.4",
-    "react-transition-group": "^4.4.5",
-    "usehooks-ts": "^3.1.1"
+    "react-transition-group": "^4.4.5"
```
akdetrick, line 122: Candidate for removal ([ticket](https://linear.app/narmi/issue/NDS-3080/remove-react-transition-group-reduce-bundle))

akdetrick:
Conflicts resolved; merging some backportable fixes before merging this PR.

PR #2143, opened 2026-07-08
fix(Chip): use theme primary border for secondary chip

akdetrick:
Closes NDS-2740

Design intended for the secondary chip to have a _themed_ border rather than a primary font color border.


akdetrick:
Author-mergable.

PR #2144, opened 2026-07-08
fix(Tooltip): fix block axis tooltip arrow position

akdetrick:
Closes NDS-2870

While the `useDropdownLayer` hook was providing the data attribute, we didn't wire up the directional styles correctly for the Tooltip arrow. This PR adds the required data attribute to raise specificity as needed to make the arrow shape work again.

**We didn't notice this until a Tooltip got within 30px of the top edge of the viewport.** The collapse/expand sidebar tooltip is fairly prominent so we noticed that one right away.

#### Test case added
Before/After
Screenshot 2026-07-08 at 6 11 25 PM
Screenshot 2026-07-08 at 6 21 07 PM


spencey-b (approved):
Works like a charm!

josh-narmi (approved):

PR #2149, opened 2026-07-09
fix(useSupportsAnchorPositioning): probe scroll-container bug lazily

chris-griffin:
`HAS_SCROLL_CONTAINER_BUG` ran a live DOM probe — create three elements, `document.body.appendChild`, measure, remove — at **module-evaluation time**, guarded only by `typeof document`. Two problems:

1. **Import can throw.** In an anchor-positioning browser, loading the bundle from a non-deferred `<head>` script hits `document.body === null` → `TypeError`, which takes down the entire module import, not just this feature.
2. **Import forces layout.** Even when it doesn't throw, the probe forces a synchronous layout as a side effect of `import`.

Swapped the eager `const` for a lazily-evaluated, cached `hasScrollContainerBug()` — the probe now runs on first real use (render time, when `body` exists) instead of at import, and its result is memoized. Also hardened the probe against a null `body`.

Export is internal (single consumer: `useAnchorPolyfill`), updated in the same change. Not re-exported from the package barrel, so no public API change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


akdetrick (approved):
Makes sense.

akdetrick:
@copilot resolve the merge conflicts in this pull request

akdetrick:
outdated. The probe is only run on specific devices

PR #2155, opened 2026-07-14
Color modes/style dictionary update

akdetrick:
Closes NDS-3103
Closes NDS-3090

This PR bumps style-dictionary handling breaking changes, does some token reorganization, and adds mode/scheme support to our build without introducing breaking changes.

See it in action here: https://60620d422ffdf100216415b2-turdfwdaab.chromatic.com/?path=/story/examples-color-modes--specimen

## Changes
- upgrade style-dictionary to v5
- update _transform_ and _format_ helpers; move them to hooks as appropriate
- reorganize existing token definitions into a structure that better matches figma
- Stub out a starter set of colors for high contrast
- Add a storybook example and toolbar tool to preview the color modes

### Output
The `dist/tokens/css/tokens.css` file has the exact same the same output as main, with these rulesets appended:

```css
@media (prefers-contrast: more) {
  :root:not([data-prefers-contrast="less"]) {
    --font-color-heading: #000000;
...
  }
}

[data-prefers-contrast="more"] {
    --font-color-heading: #000000;
 ...
}
```

## How it used to work

The `style-dictionary` library is basically a build tool that allows you to store your design tokens in declarative JSON while being able to transform values and format them for different platforms. It should be no surprise that _transforms_ and _formats_ are the most important concepts:

- "format" - a template or function that represents final file content
- "transform" - a function that can transform a name or value of a token into a more appropriate name or value for the target platform

## How it works now

Color modes should change _semantic_ tokens, not _primitive_ tokens. I changed the organization (but not the names or values) of our tokens to reflect this. "grey-200" should always be the same color, regardless of mode. It is the semantic abstractions like `font.color.primary` that will be reassigned in different color modes.

## Storybook change
Added a little tool for color modes.

https://github.com/user-attachments/assets/e7f5fa80-74ea-4eb3-93d0-32771a5fed5c


tokens/hooks/transforms.mjs
```diff
@@ -0,0 +1,74 @@
...
+    type: "name",
+    transform: ({ attributes }) => {
+      const { category, type, item } = attributes;
+      if (type === "background") return `bgColor-${item}`;
+      if (["theme", "alpha", "space", "elevation"].includes(type))
+        return `${type}-${item}`;
+      if (["narmi", "system"].includes(type)) return `${category}-${item}`;
+      return `${category}-${type}-${item}`;
+    },
+  },
```
akdetrick, line 18: These transforms have not functionally changed.

.storybook/preview.js
```diff
@@ -36,5 +36,19 @@ export const parameters = {
...
 
+export const globalTypes = {
+  contrast: {
+    description: "Color contrast mode",
+    toolbar: {
+      title: "Contrast",
+      items: [
+        { value: "", title: "Default" },
+        { value: "more", title: "High Contrast" },
+      ],
```
akdetrick, line 48: For _all_ stories, we add this toolbar item. In `decorators.js`, the state from this toolbar button is plucked from context to apply the data attribute the way we might in banking (DBBL).

src/examples/ColorModes.stories.jsx
```diff
@@ -0,0 +1,143 @@
+import React from "react";
```
akdetrick, line 1: # slop

tokens/build.ts
```diff
@@ -0,0 +1,179 @@
+/**
```
akdetrick, line 1: # not slop
The new version of style-dictionary has a much more ergonomic way of dealing with things. I never had complaints about the old API, but this one is objectively better.

This is essentially the same setup we have on `main`, but migrated for the following new conditions:

- token reorganization to support themes and modes
- style-dictionary v5 compatibility
- two instances of style-dictionary to append "mode" reassignments to generated CSS vars. Each new mode we add will get its own instance to compile to platforms.

**Transform changes are non-breaking.**

Any actual transform logic stays the same.

The artifacts generated by style-dictionary only have additive changes. Variable names and values remain as expected, with the exception of hex codes being capitalized uniformly now.

tokens/hooks/filters.mjs
```diff
@@ -0,0 +1,14 @@
+/**
+ * Custom filters for style-dictionary.
+ * Keys are filter names, values are filter functions.
+ */
+export const filters = {
+  rgbColorFilter: ({ attributes }) => {
```
akdetrick, line 6: Carried over from existing filters. This one, the `rgbColorFilter`, is used to determine which tokens should get a an entry in the partial RGB values CSS file.

tokens/hooks/formats.mjs
```diff
@@ -0,0 +1,37 @@
...
+  "custom/cjs-nested": ({ dictionary }) => {
+    const nested = buildNestedObject(dictionary.allTokens);
+    return [
+      `const tokens = ${JSON.stringify(nested, null, 2)};`,
+      "",
+      "module.exports = tokens;",
+      "",
+    ].join("\n");
+  },
+  "custom/esm-nested": ({ dictionary }) => {
```
akdetrick, line 27: One of these is used by atlas already if I'm not mistaken

tokens/build.ts
```diff
@@ -0,0 +1,183 @@
...
+    },
+    css: {
+      basePxFontSize: 16,
+      transforms: [...CSS_TRANSFORMS, "custom/value/pxToRem"],
+      buildPath: getBuildPath("css"),
+      files: [
+        {
+          destination: "tokens.css",
+          format: "custom/css-variables",
+          options: { colorScheme: false },
```
akdetrick, line 67: This allows us to include all the changes needed for dark mode support without actually releasing it yet.
Setting this to `false` prevents the format from adding the `color-scheme: light dark` rule to root styles.

tokens/hooks/transforms.mjs
akdetrick: I really should have put tests on these, but the main thing I'm concerned with is assertion on actual build artifacts in dist/

tokens/primitives/color.json
```diff
@@ -69,15 +81,30 @@
...
-      "mediumGrey": { "value": "{color.grey.mediumGrey.value}" },
-      "lightGrey": { "value": "{color.grey.lightGrey.value}" },
-      "white": { "value": "{color.grey.white.value}" }
+      "black": { "value": "{color.grey.black}" },
+      "grey": { "value": "{color.grey.mediumGrey}" },
+      "mediumGrey": { "value": "{color.grey.mediumGrey}" },
+      "lightGrey": { "value": "{color.grey.lightGrey}" },
+      "white": { "value": "{color.grey.white}" }
+    },
+    "system": {
```
wrbrand, line 94: From former `tokens/src/color/system.json`

tokens/primitives/color.json
```diff
@@ -69,15 +81,30 @@
...
+      "successDark": { "value": "#027A48" },
+      "successLight": { "value": "#ECFDF3" },
+      "warnDark": { "value": "#B54708" },
+      "warnLight": { "value": "#FEF8E3" },
+      "infoDark": { "value": "#195274" },
+      "infoLight": { "value": "#E8EEF1" },
+      "errorDark": { "value": "#b42318" },
+      "errorLight": { "value": "#FEF3F2" }
+    },
+    "alpha": {
```
wrbrand, line 104: From former `tokens/src/color/alpha.json`

tokens/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const COLOR_MODES = {
+  highContrast: {
+    mediaQuery: "(prefers-contrast: more)",
+    selector: '[data-prefers-contrast="more"]',
+    attribute: "data-prefers-contrast",
+    value: "more",
+  },
+  standardContrast: {
+    selector: '[data-prefers-contrast="no-preference"]',
```
akdetrick: **SELF** i need to change this back to "less"

akdetrick: done

wrbrand (approved):
The actual built NDS is identical here in terms of tokens if not in terms of _ordering_ (which shouldn't matter), hex case (which definitely doesn't matter), and comments, which is absolutely good enough for me.

PR #2157, opened 2026-07-16
feat(Table): add scroll behavior and pinning

akdetrick:
Closes NDS-3004

**Targeting the .8 banking release**

### Changes
- The `Row` used to control block padding. This responsibility was moved to the `Cell` which is positioned by `stretch` alignment to fit row height.
- Header cell DOM simplified
- Moved editable `kind` CSS into one place
- Added `pinnedColumns` prop to turn the start, end, or both start and end columns into sticky columns

### How to use it

`start`, `end`, or `both`

```jsx
<Table pinnedColumns="start" ... />
```

### The CSS added
The all-CSS implementation means we don't have to measure anything, listen to any events, and consider how such things are calculated or stored in the react lifecycle. 

With the Cells sitting flush against each other, we have control over backgrounds at the cell level. With that, we only need `position: sticky` on a per-cell basis using some modern, well-supported selectors. They appear as one column because they are flush.

you can't jank up the main thread with scroll events if you don't use them

https://github.com/user-attachments/assets/61ef4503-c614-40d0-94d1-7b6e93e40fe8

### DIFF
--- a/src/Table/index.scss
+++ b/src/Table/index.scss
@@ -72,7 +72,7 @@
   white-space: nowrap;
   text-overflow: ellipsis;
 
-  // Edge padding (moved from row to cell for full-bleed backgrounds)
+  // Padding for edge cells
   .nds-table-row > &:first-child,
   .nds-table-row > .nds-table-cell-wrapper:first-child & {
     padding-inline-start: rem(8px);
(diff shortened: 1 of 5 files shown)


src/Table/index.scss
```diff
@@ -56,22 +60,28 @@
...
+  // Padding for edge cells
+  .nds-table-row > &:first-child,
+  .nds-table-row > .nds-table-cell-wrapper:first-child & {
+    padding-inline-start: rem(8px);
+  }
+  .nds-table-row > &:last-child,
+  .nds-table-row > .nds-table-cell-wrapper:last-child & {
+    padding-inline-end: rem(8px);
+  }
+
```
akdetrick, line 84: This is where we moved row padding responsibility to the cells.

josh-narmi (approved):
You're a CSS wizard, Adam 🪄

PR #2165, opened 2026-07-25
chore(ts): Convert 17 files (mostly lil bits around Select/Combobox/MenuItem) to TypeScript

wrbrand:
Related to [NDS-1270](https://linear.app/narmi/issue/NDS-1270/design-system-convert-combobox-to-typescript).

None of these files need any runtime changes to convert to TypeScript, so the migration should be low-stress.

There are two bits here that are interesting, both of which I'll call out in comments. Parameter descriptions are AI-generated but seem obvious and fine. In total this should nudge us another 1.5-2% towards TypeScript by GitHub's Linguist-based measurement.


src/ContentCard/index.tsx
```diff
@@ -1,7 +1,7 @@
 import cc from "classcat";
 import React from "react";
 
-import Error from "../Error";
```
wrbrand, line 4: Shadowed the JS [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) class and should never have been allowed (@akdetrick I'd love a linter configuration that could pick this up).

src/formatters/formatDate.ts
```diff
@@ -7,16 +7,18 @@
  * formatDate(new Date('July 4, 2022'), 'short'); // '7/4/22'
  * formatDate(new Date('7/4/2022'), 'long');      // 'July 4, 2022'
  *
- * @param {Date} date native date object
- * @param {String} style formatting style (`short` or `long`)
- * @returns {String} date string formatted for display
+ * @param date native date object
+ * @param style formatting style (`short` or `long`)
```
wrbrand, line 11: The name-and-type JSDoc style seems unnecessary now that this is TS, so just name is fine.

src/formatters/formatNumber.ts
```diff
@@ -11,19 +11,31 @@
...
  *
- * @param {String|Number} input string or number to format into a number string
- * @param {String} style format style (`currency` or `percent`)
- * @returns {String} number string formatted for display
+ * @param input string or number to format into a number string
+ * @param style format style (`currency` or `percent`)
+ * @returns number string formatted for display
  */
-const formatNumber = (input, style = "currency", signDisplay = "auto", showCents = true) => {
-  let number = parseFloat(input, 10);
```
wrbrand, line 19: `parseFloat` doesn't take a second argument - someone was thinking of `parseInt`.

src/formatters/formatNumber.ts
```diff
@@ -11,19 +11,31 @@
...
+  | "always"
+  | "exceptZero";
+
+const formatNumber = (
+  input: string | number,
+  style: FormatNumberStyle = "currency",
+  signDisplay: FormatNumberSignDisplay = "auto",
+  showCents: boolean = true
+): string => {
+  const number: number = parseFloat(input as string);
```
wrbrand, line 31: `input as string` over `String(input)` because `parseFloat` already does type coercion.

src/MenuButton/MenuButtonItem.tsx
```diff
@@ -1,11 +1,22 @@
-/* eslint-disable no-unused-vars */
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
wrbrand, line 1: Switching the file extension to Typescript also switches which linter rule applies; this is necessary because `MenuButtonItem`'s constructor doesn't actually use these arguments, but they _are_ used via `child.props` by its conventional parent (`MenuButton`).

akdetrick (approved):

PR #2167, opened 2026-07-26
chore(ts): convert Pagination to TypeScript

wrbrand:
Makes `Pagination` and all its friends Typescript.

Runtime output is unchanged apart from the propTypes fix below. Verified by diffing the Babel output of the old .jsx against the new .tsx; the only deltas are the two intentional ones.

Notable details:

- totalPages was declared `PropTypes.number.isRequired` while the destructure defaulted it to 1, so rendering <Pagination /> logged a spurious "marked as required" warning. Relaxed to PropTypes.number to match both the default and the new optional type. Dev-only; propTypes are stripped from the production bundle.

- aria-current used `i === selectedIndex && "page"`, which TypeScript infers as `"" | "page"` rather than `false | "page"`, and `""` is not a valid aria-current value. Switched to a ternary yielding the exact same two runtime values (`false` or `"page"`).

- handlePageClick reads dataset off the event target, which is typed as EventTarget, so it needs an HTMLElement cast. Kept `target` rather than `currentTarget` to avoid changing behaviour.

Test file is a pure rename with no content changes (I don't think they're actually typechecked, but why not).


src/Pagination/index.stories.tsx
```diff
@@ -29,7 +29,7 @@ export const FullyControlled = () => {
         <select
           className="margin--left"
           id="totalPages"
-          onChange={(e) => setTotalPages(parseInt(e.target.value), 10)}
+          onChange={(e) => setTotalPages(parseInt(e.target.value, 10))}
```
wrbrand, line 32: `setTotalPages` is a `number` state variable defaulted to 40, so I think this was always just a typo where someone was trying to pass `10` as the second arg to `parseInt` rather than to `setTotalPages`.

src/Pagination/index.tsx
```diff
@@ -215,7 +247,7 @@ Pagination.propTypes = {
    * Total number of pages
    * If the number of pages is 1, pagination will not render
    */
-  totalPages: PropTypes.number.isRequired,
+  totalPages: PropTypes.number,
```
wrbrand, line 250: This has a default, so shouldn't be `isRequired`:
https://github.com/narmi/design_system/blob/5d35e7f491b1fb6017dc51c0d73345574b290583/src/Pagination/index.tsx#L45-L51

src/Pagination/index.stories.tsx
akdetrick: Yes

akdetrick (approved):

PR #2175, opened 2026-07-28
chore(checkbox): adds `renderLabel` prop to Checkbox for custom label elements

davidneto92:
~~Since the `label` prop is rendered within the JSX the prop type should accept it as `ReactElement` (or ReactNode, following convention with element).~~

Adds the `renderLabel` prop which will render custom elements as the label. Provides the `isChecked` state.

Screenshot 2026-08-04 at 11 07 35 AM
Screenshot 2026-08-04 at 11 07 41 AM

### DIFF
--- a/src/Checkbox/index.tsx
+++ b/src/Checkbox/index.tsx
@@ -12,7 +12,7 @@ import DisabledShim from "../DisabledShim";
 
 export interface CheckboxProps {
   /** Content of `label` element */
-  label?: React.ReactElement;
+  label?: React.ReactNode;
   /** Markdown to use in place of the `label` field */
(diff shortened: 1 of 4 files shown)


src/Checkbox/index.tsx
```diff
@@ -12,7 +12,7 @@ import DisabledShim from "../DisabledShim";
 
 interface CheckboxProps {
   /** Content of `label` element */
-  label?: string;
+  label?: React.ReactNode;
```
akdetrick: Let's go with `renderLabel` that takes a function `(isChecked: bool) => ReactNode`

akdetrick (approved):

PR #2181, opened 2026-07-29
chore(ts): convert Select/Combobox/MultiSelect item components

wrbrand:
## What

Converts the three "phantom" item components to TypeScript:

```
src/Select/SelectItem.js           -> SelectItem.tsx
src/Combobox/ComboboxItem.js       -> ComboboxItem.tsx
src/MultiSelect/MultiSelectItem.js -> MultiSelectItem.tsx
```

They render children and exist to carry props the parent introspects via `item.props.*`. Near-identical, hence one PR — following the precedent of #2165.

> Note: `MultiSelectItem` changed enough that git shows it as delete+add rather than a rename. It was moved with `git mv`; read it as a rename.

## Consumer impact: none

These are only reachable as `Select.Item`, `Combobox.Item` and `MultiSelect.Item` — statics on parents that are still `declare const` stubs in `src/index.ts`. Consumers see `any` either way until those parents are exposed.

## Two pre-existing propTypes bugs corrected

Writing the types required resolving what these props actually are, which surfaced two defects:

**1. `searchValue` was marked `.isRequired` on all three** — but every parent reads it as a fallback:

```js
item.props.searchValue || item.props.value
```

It has always been optional; the required marking produced spurious dev warnings. Now `searchValue?: string`, matching the existing correct precedent in `TableAutocomplete/Item.tsx`.

**2. `MultiSelectItem` was missing `tokenLabel` entirely** — despite `MultiSelect/index.js:71` reading `item.props.tokenLabel`, the parent's own comment documenting it, and the stories using it throughout. Added as `tokenLabel?: string`.

Both are type/propTypes-only. `value` stays required and no component body was touched, so there is **no runtime behavior change**.


src/Combobox/ComboboxItem.tsx
```diff
@@ -1,7 +1,27 @@
...
+   * the Select is open, highlight will move to the first item with a
+   * value starting with `n`.
+   */
+  value: string;
+  /**
+   * Custom typeahead string. By default, typeahead uses `value`.
+   * Use this prop to pass in a custom string you'd like the user to search
+   * against when using typeahead.
+   */
+  searchValue?: string;
```
wrbrand, line 18: Not used in this component, but used in the parent: https://github.com/narmi/design_system/blob/43d8b6fcb797a960d42ed97f827500426e59d7e1/src/Combobox/index.js#L104-L113

akdetrick (approved):

PR #2189, opened 2026-07-29
fix(types): export real types for 5 already-converted components

wrbrand:
## What

Moves `FieldToken`, `IconButton`, `LoadingShim`, `ResponsiveFlex` and `TimelineEvent` out of the `declare const` block in `src/index.ts` and into real imports.

Untyped stubs in the public API drop from **19 to 14**.

One-line fix folded in: `LoadingShim`'s `aria-busy`. Details below.

## Why

All five were converted to TypeScript in #2182–#2186. But those PRs landed as `chore(ts):` — which `release.config.js` maps to `release: false` — and **none of them updated `src/index.ts`**, so the components stayed in the untyped block.

Net result: the conversions were written, reviewed, and merged, but consumers have been receiving `any` for all five this entire time and no release ever shipped them. This is the PR that actually delivers that work.

Same operation as #2178 did for `Error`.

No runtime change. `src/index.js` is the build entry (`vite.config.js`) and already imported and exported all five, so the bundle is untouched.

## The `LoadingShim` fix

`tsconfig.json` sets `files: ["src/index.ts", "src/json.d.ts"]`, so only the entry's transitive graph is type-checked. `LoadingShim` was never in it. Pulling it in surfaced a pre-existing error:

```
src/LoadingShim/index.tsx(29,4): error TS2322:
  Types of property '"aria-busy"' are incompatible.
    Type 'string' is not assignable to type 'Booleanish'.
```

`aria-busy={isLoading.toString()}` yields `string`; React wants `boolean | "true" | "false"`. This predates the TS conversion — 81f60843 left the line untouched from the original `.js`.

Fixed as `aria-busy={isLoading}`. React stringifies booleans for `aria-*` attributes rather than omitting them, so the rendered attribute is identical. The existing assertions in `LoadingShim/index.test.js` cover both states and still pass against the rebuilt bundle.

This has to be in this PR rather than split out: `npm run build` runs `npx tsc`, so CI fails without it.

## Consumer impact

Verified against the built package with a strict consumer config. Errors are limited to two intended classes.

**Genuinely wrong values:**

```tsx
<IconButton label="x" />              // TS2741 — `name` is required
<IconButton name={someString} />      // TS2322 — not narrowed to IconName
<IconButton name="not-an-icon" />     // TS2322 — not a real icon
<TimelineEvent icon="not-an-icon" />  // TS2322
<FieldToken />                        // TS2741 — `label` is required
```

**Props that were already being silently dropped:**

```tsx
<LoadingShim className="x" />         // TS2322
<ResponsiveFlex style={{...}} />      // TS2322
<FieldToken label="t" id="nope" />    // TS2322
```

Worth flagging for anyone fixing downstream fallout: all four of these except `IconButton` have **no index signature and do not spread rest props at runtime**. So a complaint about an undeclared prop means that prop was never reaching the DOM in the first place — fix the call site, don't widen the type. `IconButton` does spread, and its `[x: string]: unknown` index signature reflects that, so arbitrary props still pass there.

Expected churn, highest first: `IconButton` (required + unioned `name`), `TimelineEvent`, `FieldToken`, then `LoadingShim` / `ResponsiveFlex` (all props optional, should be quiet).

## Follow-ups, not in this PR

- Re-export the prop interfaces (`IconButtonProps`, `FieldTokenProps`, …) from the entry. They're emitted per-file but unreachable from the package root, so consumers can't name them to write wrappers.
- The remaining 14 stubs.
- `propTypes` leaks into the emitted `.d.ts`, which is why the errors above mention `InferProps<...>`. Cosmetic, but it makes consumer error messages considerably harder to read.


src/LoadingShim/index.tsx
```diff
@@ -26,7 +26,7 @@ const LoadingShim = ({
   <div
     data-testid={testId || "nds-loadingshim"}
     aria-live="polite"
-    aria-busy={isLoading.toString()}
+    aria-busy={isLoading}
```
akdetrick, line 29: ok react

akdetrick (approved):

PR #2191, opened 2026-07-29
feat(types): export prop types from the package root

wrbrand:
## What

Exports **77 named types** from the package root: a prop type for every exported component, prop types for sub-components, and the supporting unions needed to declare values.

## Why

Consumers currently cannot name a single prop type. `src/index.ts` re-exports only `./types/Icon.types`, so the most common thing a consumer does with a design system — wrap a component — has no good answer:

```tsx
// today: ButtonProps is unnameable from the package root
const AppButton = (props: ButtonProps) => <Button {...props} kind="primary" />;
```

The options were deep-importing `@narmi/design_system/dist/types/Button` (unstable, not a supported entry point) or redeclaring props by hand (drifts immediately).

## Most of this already existed

**41 `*Props` types were already exported** from their own modules and simply unreachable from the root.

**13 more already existed with the correct name** and only lacked the `export` keyword — one-word diffs:

```diff
- interface AlertProps {
+ export interface AlertProps {
```

`Alert`, `ContentCard`, `Checkbox`, `Count`, `DisabledShim`, `Error`, `FormSection`, `Pagination`, `ProgressBar`, `Radio`, `SeparatorList`, `LoadingSkeleton`, `TruncatedAccount`.

## Two that needed real changes

1. **`Slider` exported its props as `Props`** (`src/Slider/index.tsx:6`) — far too generic to sit at a package root. Renamed to `SliderProps`. Nothing referenced it, internally or otherwise.
2. **`Snackbar` had no named props type**, only an inline `React.PropsWithChildren<{...}>`. Extracted as `SnackbarProps`, preserving the existing shape exactly.

## Naming decisions

- `Table`'s `CellProps` / `HeaderCellProps` are **aliased on the way out** to `TableCellProps` / `TableHeaderCellProps`. The bare names are too generic for a shared root namespace. Their modules still export the original names, so deep imports are unaffected.
- **Sub-component props are included** — `RowItemProps`, `TabsListProps`/`TabsPanelProps`/`TabsTabProps`, `ContextMenuItemProps`, `TableSelectItemProps`, `TableAutocompleteItemProps`, `SplitButtonMenuProps`/`SplitButtonPopoverProps`, `FieldTextProps`, `AutoCompleteProps`. Someone wrapping `Table.Cell` has exactly the same problem as someone wrapping `Table`.
- **Deliberately not exported**: `Placement`, `Alignment`, `CSSValue`, `FlexDirection`, `ViewportBreakpoint`, `ColMinBreakpoint`. Too generic to claim at the root, and hard to walk back once published. Better renamed than exported as-is — noted as a follow-up.

Verified there are **zero name collisions** across all 70 exported type names in every module reachable from the entry, and none shadow a component export.

## Risk

Purely additive. Nothing existing changes shape.

`export type` is required throughout because `isolatedModules` is on. Since `src/index.js` is the runtime entry and `index.ts` is types-only, there is no runtime impact at all.

The one theoretical break is the `Slider` rename, which would affect a consumer deep-importing `Props` from `dist/types/Slider` — a path that was never a supported entry point, for a type that was never reachable from the root.

## Verification

```tsx
import type { ButtonProps, TableCellProps, SliderProps, TabsKind, IconName } from "@narmi/design_system";
const AppButton = (props: ButtonProps) => <Button {...props} kind="primary" />;  // ✅
```

### DIFF
--- a/src/Snackbar/index.tsx
+++ b/src/Snackbar/index.tsx
@@ -52,7 +52,7 @@ export interface SnackbarProps {
   /** Content of the snackbar */
   children?: React.ReactNode;
   /** When `true`, the snackbar is rendered */
-  isActive: boolean;
+  isActive?: boolean;
 }
 
 /**
--- a/src/Snackbar/index.tsx
+++ b/src/Snackbar/index.tsx
@@ -48,16 +48,18 @@ const SnackbarButtonGroup = ({ children }: React.PropsWithChildren) => (
   </ul>
 );
 
+export interface SnackbarProps {
+  /** Content of the snackbar */
+  children?: React.ReactNode;
+  /** When `true`, the snackbar is rendered */
+  isActive: boolean;
+}
+
 /**
  * A status toolbar for multiple selection in a table.
  * Intended to be rendered in fixed position over a table.
  */
-const Snackbar = ({
-  children,
-  isActive = false,
-}: React.PropsWithChildren<{
-  isActive: boolean;
-}>) => {
+const Snackbar = ({ children, isActive = false }: SnackbarProps) => {
   return (
     <div aria-live="polite" role="status">
       {isActive && (
--- a/src/Alert/index.tsx
+++ b/src/Alert/index.tsx
@@ -5,7 +5,7 @@ import type { IconName } from "../types/Icon.types";
 
 const noop = () => {};
 
-interface AlertProps {
+export interface AlertProps {
   /** The alert is only visible when this prop is set to `true` */
   isActive?: boolean;
   /** Renders a dismiss button when `true` */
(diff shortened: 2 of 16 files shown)


src/Snackbar/index.tsx
```diff
@@ -48,16 +48,18 @@ const SnackbarButtonGroup = ({ children }: React.PropsWithChildren) => (
   </ul>
 );
 
+export interface SnackbarProps {
+  /** Content of the snackbar */
+  children?: React.ReactNode;
+  /** When `true`, the snackbar is rendered */
+  isActive?: boolean;
```
wrbrand, line 55: Changed from `isActive` to `isActive` because the component already accepts a default (`false`).

akdetrick (approved):
YES

PR #2195, opened 2026-07-29
refactor(components): align component styles with 'ink' tokens

akdetrick:
Many of our component-level styles in NDS use _primitive_ tokens instead of _semantic_ tokens.

This PR updates token usage in components to use the appropriate semantic token, which enables reassignment through forced color modes.


akdetrick:
I feel comfortable author-merging this once checks pass. This is technically a bug and these references have likely caused us pain in the past and/or will cause pain in the future

akdetrick:
Chromatic snapshot failure on Button - actually a bug fix 🙌 
We are now aligned with Figma on the SplitButton caret

PR #2197, opened 2026-07-29
fix(storybook): restore docgen for TypeScript components

wrbrand:
## Problem

Storybook shows no controls and no prop descriptions for every TypeScript component. Reproducible on the deployed site: [Toggle](https://narmi.github.io/design_system/?path=/docs/components-toggle--docs) has an empty Controls table, while [Drawer](https://narmi.github.io/design_system/?path=/docs/components-drawer--docs) (still `.jsx`) is fine.

## Root cause

The [Storybook 9 → 10 bump](https://github.com/narmi/design_system/pull/2107) pulled in `@joshwooding/vite-plugin-react-docgen-typescript@0.7.0`, which was rewritten to use the TypeScript project service. It now **skips any file that is not a root file of the resolved TypeScript project**:

```
Skipping docgen for ".../src/Toggle/index.tsx" because it is not included in
the active TypeScript project.
```

The plugin defaults to `<root>/tsconfig.json`. Ours is a build-only config — `"files": ["src/index.ts", "src/json.d.ts"]`, no `include` — so the project's root-file set is those two files, the intersection with `**/*.tsx` is empty, and **every component is skipped**.

The failure is silent: the plugin emits a `warn`, returns the source unchanged, and the build succeeds. Chromatic stayed green because the rendered stories are unaffected — only the docs metadata is gone.

`.jsx` components still work because `@storybook/react-vite` routes `.js/.jsx/.mjs` through plain `react-docgen` and only sends `.ts/.tsx` to this plugin. That's why the breakage has been growing: each `chore(ts): convert X to TypeScript` PR moves one more component from the working path onto the broken one.

Verified against the deployed bundles — `Drawer` ships full `__docgenInfo` with prop descriptions; `Toggle`, `Row`, `Sidebar` and every other `.tsx` chunk ship none.

## Changes

- **`.storybook/main.ts`** — point the plugin at `tsconfig.check.json`, which covers all of `src/`. Explicit `include: ["src/**/*.tsx"]` also avoids the plugin's default `**/*.tsx` glob walking `node_modules` on cold start.
- **`scripts/checkDocgen.mjs`** — runs the real plugin with the real options from `.storybook/main.ts` over three representative components and fails if `__docgenInfo` is missing or has no props.
- **`.github/workflows/lint.yaml`** — run the guard in the existing `run-typecheck` job.

I chose to reuse `tsconfig.check.json` rather than add another config; it already includes `src/**/*.ts(x)` and excludes tests/stories/figma files.

## Verification

- `NODE_ENV=production npm run build-storybook` — zero "not included in the active TypeScript project" warnings, and 89 emitted assets now carry `__docgenInfo`.
- Confirmed props + descriptions in the built output for TSX components (`Toggle`, `Row`, `Row.Item`, `Sidebar`, `Sidebar.Item`, `IconButton`) and that JSX components (`Drawer`, `TokenInput`) are unchanged.
- Prop counts stay sensible (`IconButton` = 11), so the plugin's default `propFilter` is still stripping `node_modules`-inherited props — no bloated control tables.
- `npm run check:docgen` passes, and fails with a clear message when `reactDocgenTypescriptOptions` is removed.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -6,6 +6,25 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 export default {
   typescript: {
     reactDocgen: "react-docgen-typescript",
+    // `@joshwooding/vite-plugin-react-docgen-typescript` (used by
+    // `@storybook/react-vite` for the `react-docgen-typescript` option) skips
+    // any file that is not a root file of the resolved TypeScript project.
+    //
+    // It defaults to the root `tsconfig.json`, which is a build-only config
+    // (`files: ["src/index.ts", "src/json.d.ts"]`). Under that config no `.tsx`
+    // file qualifies, so every TypeScript component silently loses its docgen:
+    // no controls and no prop descriptions in Storybook. Point it at
+    // `tsconfig.check.json` instead, which covers all of `src/`.
+    //
+    // `scripts/checkDocgen.mjs` guards this in CI. Do not remove these options
+    // without running `npm run check:docgen`.
+    reactDocgenTypescriptOptions: {
+      tsconfigPath: "tsconfig.check.json",
+      // Explicit `include` avoids the plugin's default `**/*.tsx` glob, which
+      // walks `node_modules` on every cold start.
+      include: ["src/**/*.tsx"],
+      exclude: ["**/*.stories.tsx", "**/*.test.tsx", "**/*.figma.tsx"],
+    },
   },
 
   stories: [
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -54,3 +54,9 @@ jobs:
       # every file in `src/` rather than just the `src/index.ts` entry graph.
       - name: tsc
         run: npm run typecheck
+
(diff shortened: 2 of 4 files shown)


.storybook/main.ts
```diff
@@ -6,6 +6,25 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
...
+    // file qualifies, so every TypeScript component silently loses its docgen:
+    // no controls and no prop descriptions in Storybook. Point it at
+    // `tsconfig.check.json` instead, which covers all of `src/`.
+    //
+    // `scripts/checkDocgen.mjs` guards this in CI. Do not remove these options
+    // without running `npm run check:docgen`.
+    reactDocgenTypescriptOptions: {
+      tsconfigPath: "tsconfig.check.json",
+      // Explicit `include` avoids the plugin's default `**/*.tsx` glob, which
+      // walks `node_modules` on every cold start.
```
elliot-wilson, line 24: 🫨

elliot-wilson (approved):
LGTM

PR #2202, opened 2026-07-31
fix(helper-classes): emit real properties from directional --none spacing utilities

wrbrand:
## Summary
- Fix a one-character Sass interpolation bug in `src/helper-classes/spacing.scss`: `#{direction}` → `#{$direction}`
- All eight directional none utilities (`margin`/`padding` × `top`/`right`/`bottom`/`left`) currently compile to an invalid declaration that browsers silently drop:
  ```css
  .margin--top--none{margin-direction:0 !important}
  ```
  (confirmed in the shipped 6.19.6 dist). They now compile to the documented `margin-top: 0 !important` etc. The axis (`--x--none`/`--y--none`) and all-sides (`--none`) variants were written longhand and were never affected.
- Adds `spacing.test.ts`, which compiles the stylesheet and asserts on real output — a misinterpolation produces valid-looking Sass, so only a compiled-CSS check can catch this class of bug.

## Impact on consumers
These classes have been silent no-ops, so **fixing them is the behavior change**. Anyone who wrote a directional `--none` class will start receiving the spacing they asked for. In the banking repo that's six source files:
- `staff/src/pages/configuration/sftp_wizard/` (ConnectivityTab, ConfigurationTab: `<p>` bottom margins disappear; SftpWizard: `<h2>` already zeroed by `.nds-typography`, likely no change)
- `staff/src/pages/institution_settings_v2/EditProductModal.tsx` (`<p>` bottom margin disappears)
- `azul/components/beneficiaries/AccountBeneficiaries.jsx` (`<td>` right padding zeroed)
- `azul/src/modules/reports/TransactionExportPage.tsx` (`<h1>` top margin, already zeroed by `.nds-typography`, likely no change)

All six are author-intended (the class was written to do exactly this). Recommend listing these call sites in the banking NDS-bump PR so reviewers eyeball those three screens. NDS `Error marginTop="none"` is unaffected (computed 0 before and after).


agreatscott (approved):
so my understanding is that this fix will result in a behavior change. it changes us to do things as advertised, but should we do a grep in our consumers to identify where a change might take place?

akdetrick (approved):

wrbrand:
> so my understanding is that this fix will result in a behavior change. it changes us to do things as advertised, but should we do a grep in our consumers to identify where a change might take place?

The grepping's already done in the PR description; only two are user-facing, one is a nothingburger, the other is a new view that should be QA'd anyway. I'll double-check when I update the banking repo, though.

PR #2212, opened 2026-08-03
refactor(Drawer): migrate to TypeScript

wrbrand:
## Summary

Migrates `Drawer` to TypeScript, following the established convention: colocated `Props` interface, PropTypes dropped, component moved from the `declare const` block into the typed import block in `src/index.ts`, and `DrawerProps` re-exported. With this, the untyped block is down to the downshift trio: Combobox, MultiSelect, Select.

## Typing decisions (runtime unchanged unless noted)

- **`children` keeps both documented shapes**: `ReactNode` or the `({ isVisible }) => ReactNode` render prop. No banking callsite uses the render prop, but it's documented and cheap to keep.
- **`onNext`/`onPrev` are plain optional `() => void`** — the old `null` default was a sentinel choice, never semantics: the disabled-arrow check was loose `== null` and React treats `null`/`undefined` onClick identically, so nothing ever distinguished them. The checks are now explicit `=== undefined` and the handlers flow straight to `onClick` with no coalescing. **Types-only breakage**: four staff wrappers (PaymentDetailDrawer, ScheduledPaymentDetailDrawer, FedNowTransactionDrawer, PaymentDetailDrawerRouter) declare and forward `(() => void) | null` handlers — verified no caller of any of them ever supplies the props, and their derive-helpers' `null` returns behave identically as `undefined`, so flipping them to plain optional in banking is mechanical with zero behavior change.
- **The outlet-portal helper** captures `getElementById` in a local, so the found-or-created element flows to `createPortal` without a non-null assertion.
- **`document && renderDrawerInOutlet()` became a ternary with `null`** — a `Document` isn't a valid JSX child type. Runtime identical: `document` is always truthy in a browser, and `false`/`null` render the same.

### DIFF
--- a/src/Drawer/index.tsx
+++ b/src/Drawer/index.tsx
@@ -23,12 +23,14 @@ export interface DrawerProps {
   onUserDismiss?: () => void;
   /**
    * Callback to handle user taking an action to go to the next element
-   * (click on the next arrow, right/down arrow key)
+   * (click on the next arrow, right/down arrow key).
+   * Pass `null` to render the next arrow disabled.
    */
   onNext?: (() => void) | null;
   /**
    * Callback to handle user taking an action to go to the previous element
-   * (click on the previous arrow, left/up arrow key)
+   * (click on the previous arrow, left/up arrow key).
+   * Pass `null` to render the previous arrow disabled.
    */
   onPrev?: (() => void) | null;
   /**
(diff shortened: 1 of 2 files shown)


src/Drawer/index.tsx
```diff
@@ -20,8 +71,8 @@ const noop = () => {};
 const Drawer = ({
   isOpen = false,
   onUserDismiss = noop,
-  onNext = null,
-  onPrev = null,
+  onNext,
+  onPrev,
```
wrbrand, line 75: Was only set to `null` to do the check at the old L113/L132; no reason we can't check `undefined` there and avoid a bunch of `?? null`s, but I can go either way here, and this is kinda just stylistic. Only one small cluster of callers actually pass `null` in practice, and those can be reworked trivially.

akdetrick (approved):

PR #2216, opened 2026-08-04
refactor(Select): migrate to TypeScript

wrbrand:
## Summary

Migrates `Select` (the largest remaining JS component, ~223 banking callsites) to TypeScript, following the established convention: colocated `Props` interface, PropTypes dropped, moved from the `declare const` block into the typed import block in `src/index.ts`, and `SelectProps`/`SelectItemProps`/`SelectActionProps`/`SelectCategoryProps` re-exported. After this, only MultiSelect and Combobox remain. 

There are a few casts here I'd love to avoid in the fullness of time, but I think getting this component typed at all is a worthy first step towards that.

## Typing decisions (runtime unchanged unless noted)

- **`id` and `label` are optional** — propTypes declared both required, but `id` has always had a fallback derived from `label` (banking omits it at **113 of 125** callsites) and `label` is omitted at 20; `DropdownTrigger.labelText` is optional.
- **The item model is a discriminated `SelectChild` union** (`ReactElement<SelectItemProps> | ReactElement<SelectActionProps>`), with `isAction` as a type guard. `SelectItemProps` gains `onSelect?: never` and `SelectActionProps` gains `value/searchValue?: never` — this makes the helpers' cross-member property access well-typed and rejects child shapes that would break `isAction`'s runtime discrimination (e.g. a `Select.Item` with an `onSelect` would be treated as an Action today).
- **`SelectActionProps.onSelect` is now required** — it was optional in the interface by mistake (propTypes says required, the reducer calls it unconditionally, all 4 banking Actions pass it).
- **downshift options typed `UseSelectProps<SelectChild | \"\">`** — the empty string is `getItemByValue`'s no-match sentinel. The `disabled` entry is **dropped from the downshift options**: `useSelect` has no such option and always ignored it (disabled behavior comes from `DropdownTrigger`). Runtime unchanged.
- Two small equivalences noted in the commit: `setUserInput(inputValue ?? \"\")`, and the selected-check compares a precomputed `selectedItemValue` (identical in all reachable states — `selectedItem` is never an action).
- `getItemIndex` stays exported (Combobox imports it); Select itself compiles clean under full `strict` (remaining strict noise in the graph is pre-existing in Row/useDropdownLayer).

## Strict-mode consumer verification (banking monorepo)

Compiled a consumer file reproducing the Select patterns from the 90 banking files importing it (125 `<Select>` / 203 subcomponent sites): mapped and conditional children, controlled + `defaultValue`, `errorText`, `Select.Action` with `clearSelectionOnAction`, categories (`kind`, `isFlat`), `searchValue`, custom `getTypeaheadString`, label omitted. Compiles clean. Negative cases correctly rejected: Item without `value`, Action without `onSelect`, Item with `onSelect`, and handlers typed narrower than `(value: string)`.

Known would-be errors in banking strict builds (all silently inert or wrong today, need a banking cleanup PR):
- **Dead props at 14 callsites**: `name` ×9 (consumers likely expect a form field name; Select never renders one), `autoComplete`, `keyHandler`, `placeholder`, `field` ×1 each — and one latent bug: `error` instead of `errorText` (RecipientSelector — the error message never displays).
- **~4 narrowed `onChange` annotations** (`(value: OperatorName)`, `(reason: Reason)`) that don't accept arbitrary strings under `strictFunctionTypes`.
- 2 dead subcomponent props (`Item.tokenLabel`, `Item.onClick`).

### DIFF
--- a/src/Select/index.tsx
+++ b/src/Select/index.tsx
@@ -27,7 +27,7 @@ export const isAction = (
 ): item is SelectActionElement => {
   let result = false;
   if (item && typeof item === "object" && "props" in item) {
-    result = "onSelect" in (item.props as object);
+    result = "onSelect" in item.props;
   }
   return result;
 };
--- a/src/Select/index.tsx
+++ b/src/Select/index.tsx
@@ -1,27 +1,33 @@
 /* eslint-disable jsx-a11y/no-noninteractive-tabindex */
 import React, { useState, useMemo } from "react";
-import PropTypes from "prop-types";
-import { useSelect } from "downshift";
+import { useSelect, UseSelectProps } from "downshift";
 import useDropdownLayer from "../hooks/useDropdownLayer";
 
 import cc from "classcat";
 import Row from "../Row";
 import DropdownTrigger from "../DropdownTrigger";
 import Error from "../Error";
-import SelectItem from "./SelectItem";
-import SelectAction from "./SelectAction";
-import SelectCategory from "./SelectCategory";
+import SelectItem, { SelectItemProps } from "./SelectItem";
+import SelectAction, { SelectActionProps } from "./SelectAction";
+import SelectCategory, { SelectCategoryProps } from "./SelectCategory";
 
 const noop = () => {};
 
+type SelectItemElement = React.ReactElement<SelectItemProps>;
+type SelectActionElement = React.ReactElement<SelectActionProps>;
+type SelectCategoryElement = React.ReactElement<SelectCategoryProps>;
+type SelectChild = SelectItemElement | SelectActionElement;
+
 /**
- * @param {Object} item a Select.Item or Select.Action component
- * @returns {Boolean} true if the item is a Select.Action
+ * @param item a Select.Item or Select.Action component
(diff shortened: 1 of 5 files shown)


src/Select/index.tsx
```diff
@@ -73,45 +84,112 @@ export const getItemIndex = (item, items) => {
...
- * @param {String} userInput most recent thing a user typed while focused on input
- * @returns {String} the string to use for typeahead for each given `selectItem`
+ * @param selectItem full react element of a select item
+ * @param userInput most recent thing a user typed while focused on input
+ * @returns the string to use for typeahead for each given `selectItem`
  */
-// eslint-disable-next-line no-unused-vars
-const defaultGetTypeAheadString = (userInput = "", selectItem) => {
+const defaultGetTypeAheadString = (
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
wrbrand, line 126: I'll be back to look over all the `// eslint-disable` usages in the repo in a separate pass.

josh-narmi (approved):

PR #2219, opened 2026-08-05
fix(color-modes): set component-level overrides for high contrast

akdetrick:
Closes NDS-3136

Adds a few component-level overrides to ensure sufficient contrast in high contrast mode.

Screenshot 2026-08-05 at 6 19 47 PM
Screenshot 2026-08-05 at 6 07 26 PM
Screenshot 2026-08-05 at 6 01 33 PM
Screenshot 2026-08-05 at 6 00 07 PM

### The `contrastMore` sass mixin

The mixin is a convenience function that outputs both the query and data attr paths for the same ruleset of style overrides.

```scss
@mixin contrastMore {
  @media (prefers-contrast: more) {
    :root:not([data-prefers-contrast="less"]) & {
      @content;
    }
  }

  [data-prefers-contrast="more"] & {
    @content;
  }
}
```

### DIFF
--- a/src/Count/index.scss
+++ b/src/Count/index.scss
@@ -10,6 +10,10 @@
   font-size: var(--font-size-s);
   line-height: 1;
 
+  @include contrastMore {
+    border: 1px solid var(--border-color-default);
+  }
+
   &--theme {
     background-color: RGBA(var(--theme-rgb-primary), var(--alpha-10));
     color: var(--theme-primary);
@@ -31,7 +35,7 @@
     color: var(--color-errorDark);
   }
   &--neutral {
-    background-color: var(--font-color-hint);
+    background-color: var(--bgColor-neutralGrey);
     color: var(--font-color-primary);
   }
 }
--- a/src/Tabs/index.scss
+++ b/src/Tabs/index.scss
@@ -177,6 +177,12 @@
         border-radius: var(--border-radius-m);
         background-color: var(--selected-bg);
         box-shadow: var(--elevation-middle);
+
+        // add a border to the pill in high contrast mode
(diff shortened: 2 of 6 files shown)


src/Tabs/index.scss
```diff
@@ -177,6 +177,12 @@
         border-radius: var(--border-radius-m);
         background-color: var(--selected-bg);
         box-shadow: var(--elevation-middle);
+
+        // add a border to the pill in high contrast mode
```
akdetrick, line 181: this looks like slop, but it's not. This mixin is being used pretty deep in a nested selector, so I wanted to be clear about what visual element is being overridden here

src/Count/index.scss
```diff
@@ -31,7 +35,7 @@
     color: var(--color-errorDark);
   }
   &--neutral {
-    background-color: var(--font-color-hint);
+    background-color: var(--bgColor-neutralGrey);
```
akdetrick, line 38: Bug fix! This should have used a `bgColor` for background to avoid readability problems in forced color modes (we darken rather than lighten font-color tokens in high contrast mode)

src/Toggle/index.scss
```diff
@@ -15,6 +15,10 @@
       left 200ms ease,
       width 200ms ease;
 
+    @include contrastMore {
+      border: 1px solid var(--border-color-default);
+    }
```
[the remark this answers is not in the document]
akdetrick: No. This is fine.

src/Count/index.scss
akdetrick: Visual snapshot change verified.

PR #2239, opened 2026-08-12
Nds 3164/android use dropdown layer

akdetrick:
Relates to NDS-3164

## Changes

**Adds test cases that reproduce the original bug and can be used to debug any future issues**

I moved the height calculation responsibility from JS to CSS. Even older browsers that need to be polyfilled in several ways support `dvh` units ([dynamic viewport units](https://www.w3.org/TR/css-values-4/#dvh)).

These units don't need to compare the full viewport to the visualViewport. They reflect whatever size the _visual viewport_ is, which is the area you can see excluding virtual keyboard. Our browser support policy actually puts us in range to safely rely on this for all supported browsers.

## Verification
Tested against a wide range of devices using the test case that originally reproduced the bugs. 

Covered paths for device/browser combos that do and do not have each of the polyfills (HAS_SCROLL_CONTAINER_BUG and the polyfill used when `useIsAnchorPositioningSupported` fails).

Screenshot 2026-08-12 at 4 21 50 PM
Screenshot 2026-08-12 at 4 21 13 PM
Screenshot 2026-08-12 at 4 20 21 PM
Screenshot 2026-08-12 at 4 19 44 PM


src/base-styles/custom-properties.scss
```diff
@@ -14,22 +14,39 @@
...
+* `calc(100dvh - anchor(...))` and `calc(anchor(top) - ...)`. `dvh`
+* shrinks when the virtual keyboard opens on Chromium/Android, giving
+* keyboard-aware sizing without JS. On the polyfill path,
+* `--js-dropdown-max-height` is still set by `useDropdownMaxHeight`.
 */
 @position-try --nds-dropdown-above {
   position-area: top;
+  max-height: calc(
+    anchor(top) - var(--space-l) - var(--nds-layer-gap, var(--space-xxs))
+  );
```
akdetrick, line 28: Every position-try rule gets a `max-height` now. This CSS rule replaces the JS hook `useDropdownMaxHeight`

src/hooks/useDropdownLayer/index.ts
```diff
@@ -68,20 +67,40 @@ export interface UseDropdownLayerOptions {
...
+    positionTryFallbacks: string;
+    margin: string;
+    positionTryOrder?: string;
+    primaryMaxHeight?: string;
+  }
 > = {
   bottom: {
     positionArea: "bottom",
     positionTryFallbacks: "--nds-dropdown-above, flip-inline",
+    positionTryOrder: "most-height",
```
akdetrick, line 92: Each placement has an individual set of instructions of what to do if initial placement will cause clipping. Today I learned that you can use a _strategy_ (the `most-height` keyword here) to programmatically determine the order of try fallbacks. 

In the future, we can probably create a single set of shared style rules for every placement, but that is definitely out of scope for the fix.

src/hooks/useDropdownLayer/index.ts
```diff
@@ -102,7 +121,13 @@ const PLACEMENT_CONFIG: Record<
...
 const useDropdownLayer = ({
   isOpen,
-  setIsOpen,
+  // `setIsOpen` remains part of the public options for API stability but
+  // is no longer read internally. `useAnchorPolyfill` previously used it
+  // to close the menu on `window.resize`; that handler was removed as
+  // part of NDS-3164 because it fired spuriously on Android soft-keyboard
+  // open. Keyboard-aware sizing is now handled in pure CSS via `dvh`
+  // + `position-try-order: most-height` on the native anchor-positioning
+  // path. Blur handling in the consuming component closes the menu.
```
akdetrick, line 130:
```suggestion
```

not helpful

src/hooks/useDropdownLayer/index.ts
```diff
@@ -161,6 +183,10 @@ const useDropdownLayer = ({
       positionAnchor: anchorName,
       positionArea,
       positionTryFallbacks,
+      ...(config.positionTryOrder && {
+        positionTryOrder: config.positionTryOrder,
+      }),
+      ...(config.primaryMaxHeight && { maxHeight: config.primaryMaxHeight }),
```
akdetrick, line 189: The order here is important. We only set a `maxHeight` inline style if the developer has specified a hard height cap

src/hooks/useDropdownLayer/useAnchorPolyfill.ts
```diff
@@ -39,30 +62,37 @@ export const calculatePosition = (
...
     layerEl.style.setProperty(
       "--js-dropdown-bottom",
-      `${vvHeight - anchorRect.top + anchorGap}px`,
+      `${window.innerHeight - anchorRect.top + anchorGap}px`,
     );
     layerEl.style.removeProperty("--js-dropdown-top");
+    layerEl.style.setProperty(
+      "--js-dropdown-max-height",
+      "calc(100vh - var(--js-dropdown-bottom, 0px) - var(--space-l))",
+    );
```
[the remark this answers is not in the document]
akdetrick: Well supported. Ok

src/hooks/useDropdownLayer/index.ts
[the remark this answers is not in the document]
akdetrick: This might actually be relevant. Looking into it.

josh-narmi (approved):

PR #2251, opened 2026-08-24
Nds 3213/tooltip zindex patch

akdetrick:
Fixes NDS-3213
Fixes NDS-3096

## Changes

- add interaction tests to core consumers of `useDropdownLayer` to check positioning via automatic chromatic snapshots
- remove z-index responsibility from `useDropdownLayer`

## Why
Many of our positioned elements use a dedicated layer to portal to (`.outlet`), which is at a z-index of 100. Some of the consumers of `useDropdownLayer` are portalled, but to `document.body` instead of `.outlet`.

This means that the .outlet forms a separate stacking context. The `Tooltip` had a 999 z-index to allow it to portal to the document body so the CSS native anchoring path could resolve the anchor correctly. However, **the hook was adding its own zindex as an inline style**, nullifying the original z-index value managed by the component.

The additional tests will help us move forward with these dropdown changes with more confidence as chromatic will force us to look at the placement of dropdown layers with our eyes now.

### Interaction tests
You can run them as a suite of unit tests, manually via storybook, and/or automatically on chromatic snapshots.

Screenshot 2026-08-24 at 1 59 40 PM

### DIFF
--- a/src/hooks/useDropdownLayer/index.ts
+++ b/src/hooks/useDropdownLayer/index.ts
@@ -204,9 +204,8 @@ const useDropdownLayer = ({
         ? anchorPositionStyles
         : polyFillLayerStyles),
 
-      // Always include display and z-index.
+      // Always include display
       display: isOpen ? "block" : "none",
-      zIndex: isPortalled ? 9 : 4,
     };
 
     return {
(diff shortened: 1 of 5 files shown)


src/hooks/useDropdownLayer/index.ts
```diff
@@ -204,9 +204,8 @@ const useDropdownLayer = ({
         ? anchorPositionStyles
         : polyFillLayerStyles),
 
-      // Always include display and z-index.
+      // Always include display
       display: isOpen ? "block" : "none",
-      zIndex: isPortalled ? 9 : 4,
```
akdetrick, line 209: Each component defines their own z-index. `isPortalled` is little more than a proxy for something the component should directly own

josh-narmi (approved):

PR #2255, opened 2026-08-27
fix(useDropdownLayer,Combobox): cap height via css

akdetrick:
Closes NDS-3176
Closes NDS-3209
Closes IMENG-3822
Closes NDS-3171
Closes NDS-3242

### Combobox
- remove the focus/blur "fix" that is no longer necessary on item selection
- set a special max height for touch devices, as the dropdown trigger here is an `input`

### useDropdownLayer
- take away all height management from JS
- set `vh` and `dvh` max heights as viewport-relative constraints instead
- clean up code we no longer need
- **add meta tag side effect:** consumers MUST include interactive-widget in the viewport meta tag.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,6 +1,9 @@
 import React from "react";
 import "./story-styles.css";
 import "../src/index.scss";
+// Injects the required viewport meta tag into the preview iframe, matching
+// what published consumers get via `src/index.ts`.
+import "../src/util/viewport";
 import { NdsStyles, ExamplesBackground } from "./decorators";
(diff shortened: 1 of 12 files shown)


josh-narmi (approved):

akdetrick:
### Merge is on hold

Why?
- i'd like to get additional features into main before this fix, so .9 can include them safely
- i'd like to take another hour or so testing edge cases against other consumers of `useDropdownLayer` within NDS

I'm confident this set of changes will resolve our issues in banking. Browserstack screenshots for gut check to come.

PR #2256, opened 2026-08-30
feat(ci): Add zizmor workflow security audit

chris-griffin:
Adds a [zizmor](https://docs.zizmor.sh/configuration/) GitHub Action to help improve security

### DIFF
--- a/.github/workflows/release-backport-on-merge.yaml
+++ b/.github/workflows/release-backport-on-merge.yaml
@@ -12,7 +12,7 @@ on:
     types: [closed]
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   compute-targets:
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,0 +1,10 @@
+# Configuration for zizmor, the GitHub Actions security auditor.
+# https://docs.zizmor.sh/configuration/
+#
+# CI runs with --min-severity high, so this ignore only covers the existing
+# high-severity finding present when the check was introduced.
+
+rules:
+  excessive-permissions:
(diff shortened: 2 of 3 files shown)


.github/workflows/release-backport-on-merge.yaml
```diff
@@ -12,7 +12,7 @@ on:
     types: [closed]
 
 permissions:
-  contents: write
+  contents: read
```
chris-griffin, line 15: This change should be safe because this only impact the built in GH_TOKEN and we really use SEMANTIC_RELEASE_TOKEN in later steps

.github/zizmor.yml
```diff
@@ -0,0 +1,11 @@
+# Configuration for zizmor, the GitHub Actions security auditor.
+# https://docs.zizmor.sh/configuration/
+
+rules:
+  artipacked:
+    ignore:
+      - release-backport.yaml:63
```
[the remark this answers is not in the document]
chris-griffin: There are other more complicated changes required with minimal benefit given the release token so ignoring for now.

.github/workflows/release-backport-on-merge.yaml
akdetrick: 👍 If a conflict happens, we fail the job with instructions. The workflow won't create the maintenance branch if it doesn't exist with RO permission, but that's probably okay. 

It's a better idea to create a separate workflow with a smaller surface area to do that ([follow up tix](https://linear.app/narmi/issue/NDS-3241/nds-automation-add-backport-create-workflow))

akdetrick (approved):

PR #2263, opened 2026-09-04
Dropdowns/patch js path

akdetrick:
Removes `HAS_SCROLL_CONTAINER_BUG` detection and forking logic. That was a stopgap solution for some Safari issues that we no longer have.

Now that our dropdowns all use `dvh`, we no longer need it. It was causing false negatives in Chrome for anchor support.

Tested against Safari (iOS 15-26.6), Firefox, Chrome (desktop and android), and Samsung browser.
With the changes in this PR, each browser follows the expected path for CSS native positioning vs polyfill. 

The polyfill path has been updated with a default z-index 1 for the sake of iOS < 26.


akdetrick:
Author-merging so we get the patch out for .9

PR #2275, opened 2026-09-11
fix(Select): keep typeahead visible and stop stale-highlight selection

agreatscott:
## Problem

Typing at a closed `Select` echoes your keystrokes in the trigger, then blanks them about half a second later. It reads as a field that discarded your input. Reported from FI UAT on a US-state dropdown (LNDG-1600).

The original triage blamed the downshift v7 bump (`2b61c7f51`) for removing select-on-type, concluding that typing never selects anything. That part doesn't hold up: downshift v9's `ToggleButtonBlur` case commits `props.items[state.highlightedIndex]`, and since `Select` forces `isOpen = true` on character input, that path is always armed. Typing does select — it commits on blur rather than per keystroke. Verified in Chromium: type `tr`, click away, "Truck" is selected.

What's actually wrong is two smaller things.

**1. The preview disappears.** downshift clears its typeahead buffer 500ms after the last keystroke. That dispatch only resets `inputValue`; `highlightedIndex` — and so the item a blur will commit — survives it. Because the trigger rendered `userInput`, the display went empty while the pending selection was still live. Native `<select>` has the same buffer but never shows it, so nothing appears to vanish.

**2. A non-matching keystroke could commit a stale item.** `getItemIndexByCharacterKey` returns the *previous* `highlightedIndex` when nothing matches, rather than `-1`. Type `t` (highlights Truck), pause past the buffer expiry, type `z` — Truck stays highlighted, and clicking away selects it. The user typed a letter matching nothing and got a selection.

## Change

- The trigger previews the item the typeahead actually resolved to, tracked in its own state so it survives the buffer expiry. The trigger now shows exactly what blurring will commit.
- A non-matching keystroke clears the highlight, so blur can't commit an item the user never typed toward. `getTypeaheadMatch` re-checks the candidate against the same typeahead string downshift matched on.
- `userInput` keeps its existing lifecycle, so `getTypeaheadString` consumers — which branch on it, as in the **Changing Typeahead Behavior** story — are unaffected.

Deliberately **not** restoring v6's per-keystroke `selectedItem`: it would fire `onChange` on every keystroke toward a match, and the end state is already correct without it.

## Scope

`MultiSelect` was checked and is unaffected — its `onStateChange` only toggles on `ToggleButtonKeyDownEnter`/`SpaceButton`/`ItemClick`, so it ignored v6's character-key selection too. `Combobox` uses `useCombobox` with a real `<input>` and is unaffected.

## Testing

9 new tests in `src/Select/index.test.js` driving the real component: preview while typing, preview surviving the buffer expiry (fake timers), commit-on-blur, no-match inertness, the stale-highlight guard, an existing selection left untouched, Escape discarding, `searchValue` matching, `Select.Action` never typeahead-selected, and the controlled `value` path. 6 of them fail on `main`.

Also verified by hand in Chromium against the Select stories, since jsdom's focus/blur simulation doesn't faithfully model the click-away path.

`src/index.test.js` fails locally for both `main` and this branch — it imports `../dist`, which needs a full `npm run build`. Unrelated.

### DIFF
--- a/src/Select/index.test.js
+++ b/src/Select/index.test.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { render, screen, fireEvent } from "@testing-library/react";
+import { render, screen, fireEvent, act } from "@testing-library/react";
 import Select, {
   isAction,
   getSelectedItemDisplay,
@@ -7,6 +7,7 @@ import Select, {
   getItemIndex,
   isHighlightedInCategory,
   isSelectedItemInCategory,
+  getTypeaheadMatch,
 } from "./";
 
 const MOCK_ITEMS = [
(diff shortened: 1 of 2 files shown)


akdetrick:
@agreatscott This has always bothered me, too. If you go look at [a native select element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/select), typeahead is supported, but we don't reflect your keystrokes. 

I added this weird behavior to update the DropdownTrigger label at the specific request of a designer that may have been confusing Select with a Combobox. Or a combobox that takes arbitrary text vs. a constrained one. The feedback we got is that "The user should see what they type but it shouldn't be an input". 

I think we now understand that to be a Combobox that validates onInputChange and rejects freeform options.

I'd actually favor removing the input reflection from Select entirely, but I'd have to run it by design.

PR #2283, opened 2026-09-18
docs(Button): restore kind control and document every kind

mitchwillman:
## NDS-3274

Two fixes to `Button`'s `kind` documentation.

### The control

`kind` lost its Storybook select in `60f2cd1` (`docs(figma): add code connect setup for button`), which replaced the inline union with the named `ButtonKind` alias. `react-docgen-typescript` reports the alias symbol, so Storybook saw an opaque `"ButtonKind"` string, could not enumerate options, and fell back to a free text input.

`ButtonKind` now derives from a `BUTTON_KINDS` const tuple. Docgen reports the expanded union again (verified by running the repo's docgen pipeline over `Button`: `kind` resolves to `"primary" | "secondary" | ... | "ai"` with `defaultValue: primary`), and the story's `argTypes` lists those same options explicitly so the control cannot silently degrade again if the type is ever rewritten as a plain alias.

### The kinds

`Button` supports seven kinds, and `tertiary` appeared in no story, doc, or description anywhere in the repo; `secondary` and `tonal` only showed up incidentally in `ButtonSizes` with no description. Added a `ButtonKinds` story rendering all seven from `BUTTON_KINDS`, and expanded the `kind` prop JSDoc so the props table describes the options.

Descriptions are deliberately visual — what each kind looks like, read off `variants.scss` — and say nothing about when to use which. Usage hierarchy is design's call, not something to infer from a stylesheet.

### Verification

- `npm run typecheck`, `npm run lint`, `npm run check:docgen`, and Button's 17 vitest tests pass
- `npx tsc` emits `ButtonKind` as the same seven-member union for consumers
- Storybook renders `kind` as a select and indexes the new `components-button--button-kinds` story

The new story adds one Chromatic baseline to accept.

### Follow-up

~10 other components carry the same latent regression (`IconButtonKind`, `RadioKind`, `TabsKind`, `CollapsibleCardKind`, `TimelineEventKind`, `ResponsiveFlexSize`/`Gap`/`Direction`, `ColMinBreakpoint`, `RadioButtonsLayouts`). Worth a separate ticket — the repo-wide alternative, `shouldExtractLiteralValuesFromEnum: true` in `.storybook/main.ts`, fixes them all at once but expands the ~340-member `IconName` union into every icon prop's table and control.


XihaoLuo (approved):
Makes sense.

